PR #4, opened 2024-09-25
feat: httpclient implementation

SKannaniOS:
## Description
This PR includes the basic setup to manage all network-related activities, along with an implementation of DispatchQueue added to the existing storage mechanism.

This PR includes the following implementation

- `HttpClientRequests` - a protocol used to declare the functions that will initiate network activities.
- `HttpClient` - a class implements the `HttpClientRequests` protocol and is responsible for preparing the `URLRequest` object to execute the network call.
- `HttpNetwork` - class performs the network call and returns either the response data or an error.

Note:
 - This PR provides a general Swift implementation related to the network requests, which will be adjusted in the future to meet the specific requirements for macOS, watchOS, and tvOS.
 
## Type of change

- [ ] Documentation update

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/Analytics/Analytics/Network/HttpClient.swift
+++ b/Analytics/Analytics/Network/HttpClient.swift
@@ -1,0 +1,116 @@
+//
+//  HttpClient.swift
+//  Analytics
+//
+//  Created by Satheesh Kannan on 24/09/24.
+//
+
+import Foundation
+
+// MARK: - HttpClientRequests
+/**
+ This protocol is designed to execute predefined network requests.
+ */
+protocol HttpClientRequests {
+    func getConfiguarationData()
+    func postBatchEvents(_ batch: String)
+}
+
+// MARK: - HttpClient
+/**
+ This class provides the implementation for the `HttpClientRequests` protocol.
+ */
+final class HttpClient {
+    let analytics: AnalyticsClient
+    
+    init(analytics: AnalyticsClient) {
+        self.analytics = analytics
+    }
+    
+    private func prepareGenericUrlRequest(for requestType: HttpClientRequestType) -> URLRequest? {
+        guard let url = self.prepareRequestUrl(for: requestType) else { return nil }
+        var urlRequest = URLRequest(url:url)
+        urlRequest.httpMethod = requestType.httpMethod
+        urlRequest.allHTTPHeaderFields = requestType.headers(analytics)
+        return urlRequest
+    }
+    
+    private func prepareRequestUrl(for requestType: HttpClientRequestType) -> URL? {
+        guard var url = URL(string: requestType.url(analytics).trimmedUrlString) else { return nil }
+        url = url.appending(path: requestType.endpoint)
+        
+        if requestType == .configuration {
+            url = url.appendQueryParameters(Constants.configQueryParams)
+        }
+        return url
+    }
+    
+    private func runRequset(_ request: URLRequest) {
+        HttpNetwork.perform(request: request) { result in
+            switch result {
+            case .success(let response):
+                print(response.toJSONString ?? "Bad response")
+            case .failure(let error):
+                print(error.localizedDescription)
+            }
+        }
+    }
+}
+
+// MARK: - HttpClientRequests
+
+extension HttpClient: HttpClientRequests {
+    func getConfiguarationData() {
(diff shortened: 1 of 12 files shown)


Analytics/Analytics/Network/HttpClient.swift
```diff
@@ -0,0 +1,116 @@
...
+        guard let urlRequest = self.prepareGenericUrlRequest(for: .configuration) else { return }
+        self.runRequset(urlRequest)
+    }
+    
+    func postBatchEvents(_ batch: String) {
+        guard var urlRequest = self.prepareGenericUrlRequest(for: .events) else { return }
+        urlRequest.httpBody = batch.utf8Data
+        
+        self.runRequset(urlRequest)
+    }
```
1abhishekpandey, line 73: Shall we consider making this function return some result?
Currently, it doesn't seem to return anything.

If you're planning to return something in another PR then can we consider adding a TODO.

Analytics/Analytics/Network/HttpClient.swift
```diff
@@ -0,0 +1,116 @@
...
+    func url(_ analytics: AnalyticsClient) -> String {
+        return switch self {
+        case .configuration: analytics.configuration.controlPlaneUrl
+        case .events: analytics.configuration.dataPlaneUrl
+        }
+    }
+    
+    var endpoint: String {
+        return switch self {
+        case .configuration: "sourceConfig"
```
1abhishekpandey, line 93: Please keep it like this:
```suggestion
        case .configuration: "/sourceConfig"
```
In Android, we are following the pattern of adding `/` in the endpoint.

Analytics/Analytics/Network/HttpClient.swift
```diff
@@ -0,0 +1,116 @@
...
+        case .events: "batch"
```
1abhishekpandey: iOS v1 uses `v1/batch` endpoint. Can we keep the same? 
(Also, Android is having the same.)

```suggestion
        case .events: "/v1/batch"
```

We can change the endPoint in future if needed to better reflect the new version of the SDK. But for now let's keep it in parity with iOS-v1 SDK.

Analytics/Analytics/Network/HttpNetwork.swift
```diff
@@ -0,0 +1,54 @@
...
+final class HttpNetwork {
+    
+    private init() {}
+    
+    private static let session: URLSession = {
+        let configuration = URLSessionConfiguration.ephemeral
+        return URLSession(configuration: configuration)
+    }()
+    
+    static func perform(request: URLRequest, completion: @escaping (Result<Data, Error>) -> Void) {
```
1abhishekpandey, line 32: Can we consider returning ENUMS in case of an Error? What's your thought on this @SKannaniOS?

We can have the same list as on Android: https://github.com/rudderlabs/kotlin-sdk-poc/blob/main/core/src/main/kotlin/com/rudderstack/core/internals/web/ErrorStatus.kt.

Analytics/Analytics/Network/HttpClient.swift
SKannaniOS: No need to add `/` as a prefix if you are appending a path component to the URL object.

Analytics/Analytics/Network/HttpNetwork.swift
SKannaniOS: For any status code other than 200, this function will return an error case containing the respective status code. The `HttpClient` class can handle this status code accordingly.

SKannaniOS: The purpose of the `HttpNetwork` class is to execute the `URLRequest` and return either the response data or an error.

1abhishekpandey (approved):

Analytics/Analytics/Network/HttpClient.swift
SKannaniOS: I used the below function call to append the url & endpoint.
Reference: https://developer.apple.com/documentation/foundation/url/3988449-appending

```
let urlString: String = "https://www.apple.com"

let endpoint = "test"

let url = URL(string: urlString).appending(path: endpoint)
```

ChryssaAliferi (approved):
👍 Nice one!!

PR #11, opened 2024-11-20
refactor: update the test cases for group event payload

SKannaniOS:
## Description

This PR updates the test cases for group event payloads.

## Type of change

- [ ] Documentation update

## Implementation Details

This PR includes multiple JSON files to serve as sample data for comparing generated models across various values, such as properties and provided options.

Note: These test cases will be updated in the future to reflect any new additions to the event payload fields.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works.


Analytics/AnalyticsTests/MockResources/Group/group_with_default_arguments.json
```diff
@@ -0,0 +1,15 @@
...
+  "groupId" : "Sample_Group_Event",
+  "integrations" : {
+    "All" : true
+  },
+  "messageId" : "<message-id>",
+  "originalTimeStamp" : "<original-timestamp>",
+  "sentAt" : "{{_RSA_DEF_SENT_AT_TS_}}",
+  "traits" : {
+    "anonymousId" : "<anonymous-id>"
+  },
```
1abhishekpandey: This is an incorrect behaviour. Group events shouldn't have default traits in the root. If the user has provided the value then add anonymousId field else this root.traits shouldn't be present.

Please verify the behaviour from iOS-v1 SDK. (Although, I've already verified it.)

Refer to the Android [test](https://github.com/rudderlabs/kotlin-sdk-poc/blob/develop.poc/core/src/test/resources/message/group/group_with_default_arguments.json) case for more details.

1abhishekpandey (changes requested):

Analytics/AnalyticsTests/MockResources/Group/group_with_default_arguments.json
SKannaniOS: I agree & I'll make the change. But please confirm are we adding anonymousId in the root.traits.? In V1 iOS we are adding anonymousId into context.traits. Please confirm.

SKannaniOS: Also either user provided traits or not, context.traits is added with the anonymousId. Please confirm.

1abhishekpandey: > But please confirm are we adding anonymousId in the root.traits.? In V1 iOS we are adding anonymousId into context.traits. Please confirm.

We need to add `anonymousId` in `root.traits` if, the user passes any group traits, else we don't need to have a default group traits.

> Also either user provided traits or not, context.traits is added with the anonymousId. Please confirm.

Yes, correct. Context.traits always has the anonymousId (irrespective of the user has passed any specific traits or not.)

This behaviour aligns with the v1 and other SDKs.

1abhishekpandey: After discussing with the team, we are aligned that `group` event will not contain any `anonymousId` field at the `root.traits`, if this feature is requested later on, we'll add this.

The update behaviour:
- In case nothing is passed as group traits, then `root.traits` should not be set in the payload.
- If group traits are passed then `root.traits` should contain only those passed traits.

NOTE: context.traits will continue to have `anonymousId` field. (This field needs to be implemented along with `identify` event.)

CC: @ChryssaAliferi @SKannaniOS.

1abhishekpandey: Update: [Here](https://linear.app/rudderstack/issue/SDK-2676/[android]-remove-anonymousid-field-from-roottraits-in-group-events) is the card to revert this feature on Android.

1abhishekpandey (approved):

ChryssaAliferi (approved):

PR #49, opened 2025-03-20
feat: shut down api implementation

SKannaniOS:
## Description

This PR adds shut down functionality in analytics.

## Type of change

- [ ] Documentation update

## Implementation Details

Shutdown follows this sequence of actions:
- Sets `isAnalyticsShutdown` to `true`, immediately blocking all public APIs from returning valid results and logging an error message if any public API is called.
- Closes the `ProcessMessageChannel`.
- Once all buffered events in the `ProcessMessageChannel` are processed, it invokes `teardown` for all plugins (if implemented).
- Dismantles the `PluginChain`, which also stops the `EventManager`, closing the `writeChannel` and canceling the `uploadChannel`.
- Tears down the `LifecycleSessionWrapper`, releasing the `LifecycleObserver` and `SessionHandler`, ensuring all observers are detached.

## Checklist

- [ ] I have checked for potential performance impacts and optimized if necessary.

## How to test?

Send some events, use the `shutdown` in the sample app and then try and send some events. They will fail with an error message.

## Maintainers Checklist

- [ ] CI tests have passed.

image

## Additional Context

- For now, `deinit` is used, and no `tearDown` method is implemented to release the plugin’s resources. If needed `tearDown` method will be implemented in future.

### DIFF
--- a/Analytics/Analytics/LifecycleManagement/LifecycleManagementUtils.swift
+++ b/Analytics/Analytics/LifecycleManagement/LifecycleManagementUtils.swift
@@ -89,6 +89,7 @@ class LifecycleSessionWrapper {
     }
     
     func tearDown() {
+        // Don't change the order..
         self.sessionHandler = nil
         self.lifecycleObserver = nil
     }
--- a/Analytics/Analytics/LifecycleManagement/LifecycleManagementUtils.swift
+++ b/Analytics/Analytics/LifecycleManagement/LifecycleManagementUtils.swift
@@ -80,12 +80,17 @@ class WeakObserver {
  A wrapper that integrates lifecycle observation and session handling for analytics tracking.
  */
 class LifecycleSessionWrapper {
-    var lifecycleObserver: LifecycleObserver
-    var sessionHandler: SessionHandler
+    var lifecycleObserver: LifecycleObserver?
+    var sessionHandler: SessionHandler?
     
     init(analytics: AnalyticsClient) {
         self.lifecycleObserver = LifecycleObserver(analytics: analytics)
         self.sessionHandler = SessionHandler(analytics: analytics)
+    }
+    
+    func tearDown() {
+        self.sessionHandler = nil
+        self.lifecycleObserver = nil
     }
 }
 
--- a/AnalyticsApp/AnalyticsApp/ContentView.swift
+++ b/AnalyticsApp/AnalyticsApp/ContentView.swift
@@ -103,6 +103,10 @@ struct ContentView: View {
                 CustomButton(title: "Shutdown") {
                     AnalyticsManager.shared.shutdown()
                 }
+                
+                CustomButton(title: "Initialize SDK") {
+                    AnalyticsManager.shared.initializeAnalyticsSDK()
+                }
             }
         }
     }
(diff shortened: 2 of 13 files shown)


vgupta98 (approved):

Analytics/Analytics/LifecycleManagement/LifecycleManagementUtils.swift
```diff
@@ -80,13 +80,18 @@ class WeakObserver {
...
     
     init(analytics: AnalyticsClient) {
         self.lifecycleObserver = LifecycleObserver(analytics: analytics)
         self.sessionHandler = SessionHandler(analytics: analytics)
     }
+    
+    func tearDown() {
+        self.sessionHandler = nil
+        self.lifecycleObserver = nil
+    }
```
1abhishekpandey: I tested and found this is not working for `lifecycleObserver` i.e., `lifecycleObserver` `deinit` block is not being called and I can still see the lifecycle callbacks being called. I waited for quite a long time still `deinit` block is not being called.

Steps to reproduce:
1. Run the app.
2. Call shutdown.
3. Add a debug pointer over [here](https://github.com/rudderlabs/swift-sdk-poc/blob/64c1b5cd2e8575ac1fda7279d241e5adb0b77810/Analytics/Analytics/LifecycleManagement/LifecycleObserver.swift#L52).
4. Move the app to the background/foreground. I can see the lifecycle observers are still running.

1abhishekpandey: I would strongly recommend using plugin architecture for these two plugins if there's a slight probability that `deinit` will not be called instantly. We need to ensure, on `shutdown` the state of the SDK is as expected (in this case we need to remove the observers).

AnalyticsApp/AnalyticsApp/ContentView.swift
```diff
@@ -98,6 +98,12 @@ struct ContentView: View {
                     AnalyticsManager.shared.endSession()
                 }
             }
+            
+            HStack {
+                CustomButton(title: "Shutdown") {
+                    AnalyticsManager.shared.shutdown()
+                }
+            }
```
1abhishekpandey: Let's add a button to also initialise the SDK.

1abhishekpandey (changes requested):

Analytics/Analytics/LifecycleManagement/LifecycleManagementUtils.swift
SKannaniOS: This issue is related to the retain cycle. I fixed it.

1abhishekpandey: It’s fine for now, but I’m concerned about the current approach. What if someone unintentionally **modifies the retain cycle** in the future? The issue could resurface, and there’s no unit test to catch this scenario.

Can we confidently say that the current approach prevents this mistake from happening again?

SKannaniOS: Yes, I’m confident in the current approach. Even with a plugin architecture, we still need to check for any remaining active references at some point, especially since we are observing the app’s lifecycle events.

1abhishekpandey: Let's internally document the process of checking remaining active references specifying when we need to run this check and the process.

1abhishekpandey (approved):

ChryssaAliferi (approved):

PR #113, opened 2025-07-08
fix: allow zero milliseconds for session timeout in Objective-C configuration

SKannaniOS:
## Description
This PR fixes the session timeout validation logic in the Objective-C session configuration builder to accept zero (0) milliseconds as a valid timeout value. Previously, the validation only accepted positive values (`> 0`), but now it accepts zero and positive values (`>= 0`) to support immediate session timeout scenarios.

## Type of change
- [ ] Documentation update

## Implementation Details
- Updated the validation condition in `setSessionTimeoutInMillis(_:)` from `timeoutInMillis.int64Value > 0` to `timeoutInMillis.int64Value >= 0`
- Enhanced the method documentation to explicitly state that "Zero(0) indicates immediate timeout"
- Maintains backward compatibility while enabling new use case for immediate session timeout

## Checklist

## How to test?
1. Test with positive timeout values (existing functionality should work as before)
2. Test with zero timeout value - should now be accepted and stored correctly
3. Test with negative values - should still be rejected (no change in behavior)
4. Verify that setting timeout to 0 results in immediate session timeout behavior

**Test case example:**
```objc
ObjCSessionConfigurationBuilder *builder = [[ObjCSessionConfigurationBuilder alloc] init];
[builder setSessionTimeoutInMillis: @0]; // Should now work
SessionConfiguration *config = [builder build];
// Verify that sessionTimeoutInMillis is set to 0
```

## Breaking Changes
None. This is a backward-compatible change that only expands the accepted range of values.

## Maintainers Checklist
- [ ] CI tests have passed.

N/A - This is a logic/API change with no UI components.

## Additional Context
- This change affects only the Objective-C configuration builder; the underlying SessionConfiguration already supports UInt64 values including 0
- The change enables developers to implement immediate session timeout scenarios where sessions should end as soon as the app goes to background

### DIFF
--- a/Sources/RudderStackAnalytics/ObjC/ObjCSessionConfiguration.swift
+++ b/Sources/RudderStackAnalytics/ObjC/ObjCSessionConfiguration.swift
@@ -49,7 +49,7 @@ public final class ObjCSessionConfigurationBuilder: NSObject {
     /**
      Sets the session timeout duration in milliseconds.
 
-     - Parameter timeoutInMillis: A positive number representing the timeout duration.
+     - Parameter timeoutInMillis: A positive number representing the timeout duration. Zero(0) indicates immediate timeout.
      - Returns: The builder instance for chaining.
      */
     @objc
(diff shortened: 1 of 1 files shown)


1abhishekpandey (approved):

ChryssaAliferi (approved):

ChryssaAliferi:
I am thinking, should we add a test case to make sure this change in the value is always tested?

SKannaniOS:
> I am thinking, should we add a test case to make sure this change in the value is always tested?

This is an Objective-C wrapper for the underlying Swift method. I’ll review this scenario and add test cases in future PRs if necessary.

PR #114, opened 2025-07-09
fix: resolve shutdown implementation issues

SKannaniOS:
## Description
This PR fixes the shutdown issue in the Swift SDK by implementing a proper graceful shutdown mechanism. The previous implementation had race conditions and resource cleanup issues that could lead to memory leaks and incomplete event processing during shutdown. 

The solution replaces the DispatchQueue-based event processing with AsyncChannel-based asynchronous tasks and implements proper shutdown hooks to ensure all resources are cleaned up correctly.

## Type of change
- [ ] Documentation update

## Implementation Details
- **Replaced DispatchQueue with AsyncChannel**: Migrated from `eventProcessingQueue` to `processEventChannel` for better async/await support and graceful shutdown handling
- **Added shutdown hooks**: Implemented `shutdownHook()` method that properly cleans up data plane plugins, plugin chains, and lifecycle components
- **Enhanced EventManager shutdown**: Added `isShuttingDown` flag and proper task coordination to prevent race conditions during shutdown
- **Improved AsyncChannel**: Added termination handlers, proper error handling, and thread-safe operations
- **Optimized DateFormatter**: Made ISO8601DateFormatter static to improve performance and prevent repeated instantiation

## Checklist

## How to test?
The updated test `test_shutdown()` demonstrates the fix:

1. **Initialize Analytics client** and verify it's active
2. **Call shutdown()** and wait for graceful completion (300ms)
3. **Verify shutdown state**: 
   - `isAnalyticsActive` should be `false`
   - `lifecycleObserver` and `sessionHandler` should be `nil`
4. **Attempt to send events** after shutdown
5. **Verify no events are processed** - storage should remain empty
6. **Verify all user properties are nil** after shutdown

**Test Environment:**
- Run existing unit tests to ensure no regressions
- Execute the enhanced `test_shutdown()` test specifically
- Test under various shutdown scenarios (immediate shutdown, shutdown during event processing, etc.)

## Breaking Changes
None. This is a backward-compatible bug fix that improves the internal shutdown mechanism without changing the public API.

## Maintainers Checklist
- [ ] CI tests have passed.

N/A - This is an internal API improvement with no UI changes.

## Additional Context
This fix addresses a critical issue where the SDK's shutdown process was not properly coordinating between different asynchronous components, leading to:

1. **Memory leaks** due to incomplete cleanup of plugin chains and lifecycle observers
2. **Race conditions** during event processing shutdown
3. **Incomplete event processing** when shutdown occurred during active event handling

The new implementation ensures:
- **Graceful shutdown** with proper task coordination
- **Complete resource cleanup** through the shutdown hook mechanism  
- **Thread-safe operations** with proper locking and channel state management

**Testing Notes:** The solution maintains backward compatibility while significantly improving shutdown reliability. All existing functionality remains unchanged from a public API perspective.


Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
```diff
@@ -128,7 +128,8 @@ extension RudderStackAnalyticsTests {
     func test_shutdown() async {
         guard let client = analytics_disk else { XCTFail("No disk client"); return }
         client.shutdown()
-
+        try? await Task.sleep(nanoseconds: 300_000_000)
```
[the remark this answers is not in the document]
SKannaniOS: All test cases will be revamped in future.

Sources/RudderStackAnalytics/Helpers/Extensions.swift
```diff
@@ -314,6 +311,25 @@ extension Data {
...
+     Waits synchronously for the task to complete.
+     
+     This method blocks the current thread until the task finishes execution.
+     Useful for graceful shutdown scenarios where you need to wait for async tasks to complete.
+     */
+    func waitForCompletion() {
+        let group = DispatchGroup()
+        group.enter()
+        Task {
+            _ = await self.value
```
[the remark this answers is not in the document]
SKannaniOS: The existing implementation provides the expected behaviour. So no need to improve it.

Sources/RudderStackAnalytics/EventManagement/EventUploader.swift
```diff
@@ -0,0 +1,94 @@
...
+                        
+            for await _ in self.uploadChannel.receive() {
+                // If shutdown is initiated, don't start new upload cycles
+                if self.isShuttingDown { break }
+                
+                // Read all available batched events from storage
+                let dataItems = await self.storage.read().dataItems
+                for item in dataItems {
+                    // Check shutdown conditions before processing each item
+                    if self.analytics.isAnalyticsShutdown || self.isShuttingDown { break }
```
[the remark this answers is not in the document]
SKannaniOS: No need for now.

Sources/RudderStackAnalytics/Base/Analytics.swift
```diff
@@ -310,15 +349,45 @@ extension Analytics {
...
+    private func startProcessingEvents() {
+        
+        self.isAnalyticsShutdown = false
+        
+        self.processEventChannel.setTerminationHandler { [weak self] in
+            // Only cancel immediately if not shutting down gracefully
+            if self?.isAnalyticsShutdown != true {
+                self?.processEventTask?.cancel()
+            }
+            self?.processEventTask = nil
```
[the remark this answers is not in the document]
SKannaniOS: We will analyze it & fix it in future PRs.

Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
```diff
@@ -127,12 +127,26 @@ extension RudderStackAnalyticsTests {
...
+        // Verify analytics is initially active
+        XCTAssertTrue(client.isAnalyticsActive)
+        XCTAssertNotNil(client.lifecycleObserver)
+        XCTAssertNotNil(client.sessionHandler)
+        
+        // Shutdown the analytics client
         client.shutdown()
+        
+        // Give some time for graceful shutdown to complete
+        try? await Task.sleep(nanoseconds: 300_000_000)
```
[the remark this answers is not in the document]
SKannaniOS: Test cases will be revamped in future PRs.

Sources/RudderStackAnalytics/Helpers/Extensions.swift
```diff
@@ -314,6 +314,25 @@ extension Data {
...
+            group.leave()
+        }
+        group.wait()
```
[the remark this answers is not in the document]
SKannaniOS: In whole implementation, this method is not called at Main thread. So there will be no deadlock.

Sources/RudderStackAnalytics/Base/Analytics.swift
```diff
@@ -310,15 +349,45 @@ extension Analytics {
...
     /**
-     Processes an event by dispatching it to the serial queue to ensure ordering without blocking the main thread.
+     Starts the background task that processes events from the channel through the plugin chain.
+     
+     This method creates a background task that continuously listens for events and processes them.
+     The task handles graceful shutdown and ensures proper cleanup when completed.
+     */
+    private func startProcessingEvents() {
+        
+        self.isAnalyticsShutdown = false
```
1abhishekpandey: As discussed offline, this might not be needed. @SKannaniOS will examine this further and remove it if not required.

Sources/RudderStackAnalytics/Base/Analytics.swift
```diff
@@ -310,15 +349,45 @@ extension Analytics {
...
+        
+        self.processEventChannel.setTerminationHandler { [weak self] in
+            // Only cancel immediately if not shutting down gracefully
+            if self?.isAnalyticsShutdown != true {
+                self?.processEventTask?.cancel()
+            }
+            self?.processEventTask = nil
+        }
```
1abhishekpandey: As this is not directly related to the processEvents method, let's extract it out.

Sources/RudderStackAnalytics/Base/Analytics.swift
```diff
@@ -257,6 +263,38 @@ extension Analytics {
...
+    private func shutdownHook() {
+        
+        var dataPlanePlugin: RudderStackDataPlanePlugin?
+        self.pluginChain?.apply { plugin in
+            if plugin is RudderStackDataPlanePlugin {
+                dataPlanePlugin = plugin as? RudderStackDataPlanePlugin
+            }
+        }
+        
+        dataPlanePlugin?.shutdown()
```
1abhishekpandey: As discussed offline, let's plan to refactor this code to make it scalable and ensure it fully adheres to the open-closed principle, aligning with the Kotlin implementation.

@SKannaniOS is going to work on this refactor, and after that, we will review it again.

Sources/RudderStackAnalytics/Plugins/RudderStackDataPlanePlugin.swift
1abhishekpandey, line 1: We are calling the `eventManager.stop` method twice, once when shutdown API is called and another when `deinit` block is called. 

I believe we should remove the `deinit` block completely and rely on the manual shutdown API invocation:
```
deinit {
    self.eventManager?.stop()
    self.eventManager = nil
}
```

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -48,35 +55,78 @@ extension EventManager {
...
-        self.uploadChannel.close()
+        // Close upload channel first to signal uploader to stop receiving
+        if !self.uploadChannel.isClosed {
+            self.uploadChannel.close()
+        }
+        
+        // Wait for upload task to complete naturally
+        if let task = self.uploadEventTask {
+            task.waitForCompletion()
+        }
```
1abhishekpandey: It seems a two-step process to close the `upload` channel. Let's extract them out to a new method which should have both the steps and call that single method.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -48,35 +55,78 @@ extension EventManager {
...
+        
+        // Close write channel to signal processor to stop receiving
+        if !self.writeChannel.isClosed {
+            self.writeChannel.close()
+        }
+        
+        // Wait for write task to complete any remaining events
+        if let task = self.writeEventTask {
+            task.waitForCompletion()
+        }
```
1abhishekpandey: It seems a two-step process to close the `write` channel. Let's extract them out to a new method which should have both the steps and call that single method.

Sources/RudderStackAnalytics/Base/EventManager.swift
1abhishekpandey: This will improve readability a bit.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -8,16 +8,25 @@
...
     private let writeChannel: AsyncChannel<ProcessingEvent>
     private let uploadChannel: AsyncChannel<String>
-    private let flushEvent = ProcessingEvent(type: .flush)
+    
+    /* Task management */
+    private var writeEventTask: Task<Void, Never>?
+    private var uploadEventTask: Task<Void, Never>?
+    
+    /* Flag to prevent multiple shutdown calls and coordinate cleanup */
+    private var isShuttingDown = false
```
1abhishekpandey: Can we carefully examine if this is really needed or not? 

Try to explore if we can rely upon `analytics.isAnalyticsShutdown` variable or not?

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -48,35 +55,78 @@ extension EventManager {
...
+    
+    // MARK: - Event Processing
+    private func write() {
+        self.writeChannel.setTerminationHandler { [weak self] in
+            // Only cancel immediately if not shutting down gracefully
+            if self?.isShuttingDown != true {
+                self?.writeEventTask?.cancel()
+            }
+            self?.writeEventTask = nil
+        }
```
1abhishekpandey: Extract it to a separate method.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -98,22 +153,40 @@ extension EventManager {
...
-        Task { [weak self] in
+    // MARK: - Event Uploading
+    private func upload() {
+        self.uploadChannel.setTerminationHandler { [weak self] in
+            // Only cancel immediately if not shutting down gracefully
+            if self?.isShuttingDown != true {
+                self?.uploadEventTask?.cancel()
+            }
+            self?.uploadEventTask = nil
+        }
```
1abhishekpandey: Extract it to a separate method.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -126,7 +199,11 @@ extension EventManager {
 }
 
 // MARK: - ProcessingEvent
-private class ProcessingEvent {
+/**
+ * ProcessingEvent represents an event in the processing pipeline.
+ * It can be either a regular message event or a flush signal.
+ */
+class ProcessingEvent {
```
1abhishekpandey: Why did we removed `private` modifier? Let's add it back if we don't intend to use this outside of this class.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -48,35 +55,78 @@ extension EventManager {
...
+        }
+        
+        // Close write channel to signal processor to stop receiving
+        if !self.writeChannel.isClosed {
+            self.writeChannel.close()
+        }
+        
+        // Wait for write task to complete any remaining events
+        if let task = self.writeEventTask {
+            task.waitForCompletion()
```
1abhishekpandey: Also, during debugging, the control never reached over here. Can you please check if it's working as expected or not?

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -137,7 +214,10 @@ private class ProcessingEvent {
 }
 
 // MARK: - ProcessingEventType
-private enum ProcessingEventType {
+/**
+ * Enum representing the different types of processing events.
+ */
+enum ProcessingEventType {
```
1abhishekpandey: Why did we removed `private` modifier? Let's add it back if we don't intend to use this outside of this class.

Sources/RudderStackAnalytics/Base/Analytics.swift
SKannaniOS: Removed.

Sources/RudderStackAnalytics/Base/Analytics.swift
SKannaniOS: Modified the implementation to be simpler.

Sources/RudderStackAnalytics/Base/Analytics.swift
SKannaniOS: This section is removed. The shutdown implementation is now more simpler.

Sources/RudderStackAnalytics/Plugins/RudderStackDataPlanePlugin.swift
SKannaniOS: Refactored the shutdown process to be more straightforward.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: Refactored the shutdown process to be more straightforward.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: Refactored the shutdown process to be more straightforward.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: Refactored the shutdown process to be more straightforward.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: Refactored the shutdown process to be more straightforward.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: Refactored the shutdown process to be more straightforward.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -48,35 +52,60 @@ extension EventManager {
...
-}
-
-// MARK: - Event Processing
-extension EventManager {
-    func write() {
-        Task { [weak self] in
+    
+    // MARK: - Event Processing
+    private func write() {
+
```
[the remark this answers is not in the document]
SKannaniOS: Done.

Sources/RudderStackAnalytics/Helpers/AsyncChannel.swift
```diff
@@ -9,47 +9,96 @@ import Foundation
...
+    private var continuation: AsyncStream<Element>.Continuation?
+    private let stream: AsyncStream<Element>
     private let lock = NSLock()
     
+    /**
+     Indicates whether the channel is closed.
+     */
+    var isClosed = false
+    
+    /**
```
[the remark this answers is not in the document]
SKannaniOS: Done.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -98,22 +121,32 @@ extension EventManager {
...
-    func upload() {
+    // MARK: - Event Uploading
+    private func upload() {
         Task { [weak self] in
             guard let self else { return }
             
-            for await _ in self.uploadChannel.stream {
+            for await _ in self.uploadChannel.receive() {
+                // If shutdown is initiated, don't start new upload cycles
+                if self.analytics.isAnalyticsShutdown { break }
```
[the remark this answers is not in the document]
SKannaniOS: Both serve the same purpose and act as aliases for each other. So no need to change it.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -98,22 +121,32 @@ extension EventManager {
...
+                
+                // Read all available batched events from storage
                 let dataItems = await self.storage.read().dataItems
                 for item in dataItems {
-                    LoggerAnalytics.debug(log: "Upload started: \(item.reference)")
+                    // Check shutdown conditions before processing each item
+                    if self.analytics.isAnalyticsShutdown { break }
```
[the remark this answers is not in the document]
SKannaniOS: Both serve the same purpose and act as aliases for each other. So need to change it.

Sources/RudderStackAnalytics/Helpers/AsyncChannel.swift
```diff
@@ -9,47 +9,96 @@ import Foundation
...
+     Sets a handler to be called when the channel terminates.
+     
+     - Parameter completion: The closure to call when the channel terminates.
+     */
+    func setTerminationHandler(_ completion: (() -> Void)?) {
+        continuation?.onTermination = { @Sendable _ in
+            completion?()
+        }
+    }
```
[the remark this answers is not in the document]
SKannaniOS: This implementation will be used if needed in future. So no need to remove it.

Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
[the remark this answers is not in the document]
SKannaniOS: All test cases will be revamped in future.

Sources/RudderStackAnalytics/Base/EventManager.swift
1abhishekpandey, line 131: Having one shutdown check is enough. Let's remove this check and keep the next check.
```suggestion
```

Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
1abhishekpandey, line 140: We shouldn't use any real delay in unit test case. Please remove the sleep statement and consider other ways to verify the test if possible. 

Adding delays will extend the time to execute unit tests. Ideally, unit tests should be fast enough to execute asap.

1abhishekpandey: If sleep statement is used somewhere else please consider removing it.

Sources/RudderStackAnalytics/Base/Analytics.swift
```diff
@@ -310,15 +321,34 @@ extension Analytics {
...
+     Starts the event processing task that handles events from the channel.
+     
+     This method creates a concurrent task that continuously processes events through the plugin chain.
+     The task includes automatic cleanup via defer block to ensure proper resource management.
+     */
+    private func startProcessingEvents() {
+        Task { [weak self] in
+            guard let self else { return }
+            
+            defer { self.shutdownHook() }
```
1abhishekpandey, line 333: NO ACTION NEEDED FOR NOW.

In `shutdownHook()`, we've this code:
```
self.lifecycleSessionWrapper?.tearDown()
self.lifecycleSessionWrapper = nil
```

From now on, let's ensure we don't use any wrapper class; instead, try to extend the plugin class. Using a plugin helps us utilise the true power of plugin architecture, as we can override the `teardown` method directly.

1abhishekpandey (commented):
Great work!

Sources/RudderStackAnalytics/Helpers/AsyncChannel.swift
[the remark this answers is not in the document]
1abhishekpandey: Let's not write code for the future. In the future, needs might change. Let's remove it for now and add it whenever required.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: No. The shutdown check should be placed in two locations:

1. Outer check — Prevents reading any new batches for upload after shutdown is initiated.
2. Inner check — Prevents processing the next batch that’s already being prepared or is about to be uploaded.

Each one serves as a fallback mechanism for the other.

Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
SKannaniOS: As previously discussed, there will be a complete revamp of all test cases under the related ticket. All artificial delays will be removed as part of that effort.

Sources/RudderStackAnalytics/Base/Analytics.swift
SKannaniOS: Just to clarify, this wrapper class holds references to both `LifecycleObserver` (which observes lifecycle changes) and `SessionHandler` (which handles session-related logic). I plan to convert these classes into plugins in the future.

Sources/RudderStackAnalytics/Base/EventManager.swift
1abhishekpandey: > Outer check — Prevents reading any new batches for upload after shutdown is initiated.

Once the shutdown is initiated, the flush logic will be stopped.

I believe an outer check is not really needed for these reasons:
```
if self.analytics.isAnalyticsShutdown { break }         // < - Outer check
// Read all available batched events from storage
let dataItems = await self.storage.read().dataItems
for item in dataItems {
  // Check shutdown conditions before processing each item     
  if self.analytics.isAnalyticsShutdown { break }     // <- Inner check.
```
- Let's consider we don't have an outer check, then at worst, the data will be read once, but the SDK will not attempt to send it. The control will still return when the inner check runs.
- The distance between the outer and inner check is just two lines away.
- Performance-wise wise the impact is negligible, and our business logic is also intact, even if we have just one inner check.
- Also, having one check simplifies the code.

Let me know your thoughts.

Sources/RudderStackAnalytics/Base/Analytics.swift
1abhishekpandey: Great.

Sources/RudderStackAnalytics/Base/EventManager.swift
SKannaniOS: While the code only appears to have two lines, invoking `storage.read()` triggers an `asynchronous` file-read operation that can delay the overall flow. By keeping both checks in place, we can avoid this delay by preventing unnecessary reads upfront.

1abhishekpandey: I've internally proposed a way to optimise this batch read code further and if we do that, we don't need this outer check as the impact will be reduced a lot.

If we align on this, we can implement the changes in a new PR. For now, I'll resolve this comment.

Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
1abhishekpandey: Sure, but let's not write new code with having `sleep` statement.

1abhishekpandey (approved):

ChryssaAliferi (approved):

PR #116, opened 2025-08-14
refactor: optimize event batch upload logic

SKannaniOS:
## Description
This PR refactors the event batch upload logic to improve performance and memory efficiency. The key optimization is to avoid loading large batch content into memory when creating `EventDataItem` objects during retrieval, and instead read the file content only when needed during the upload process.

The refactor addresses potential memory issues where all event batches would be loaded into memory simultaneously during storage retrieval, which could cause performance problems with large batches or when multiple batches are queued for upload.

## Type of change
- [ ] Documentation update

## Implementation Details
- **EventManager.swift**: Implemented lazy loading for disk storage mode - batch content is now read from file only when needed for upload, not during retrieval
- **DiskStore.swift**: Optimized retrieval to only check file existence instead of loading content into memory
- **StorageModels.swift**: Made `EventDataItem` batch parameter optional with default empty value
- **Tests**: Updated test assertions to handle the new lazy-loading approach and added proper cleanup

## Checklist
- [ ] I have added the necessary documentation (if appropriate).

## How to test?
1. **Performance Testing**: Run the existing test suite to ensure functionality remains intact
2. **Memory Testing**: Monitor memory usage when processing large batches

## Breaking Changes
None. This is a performance optimization that maintains the same public API and functionality.

## Maintainers Checklist
- [ ] CI tests have passed.

N/A - This is a performance optimization with no UI changes.

## Additional Context
- **Performance Impact**: Reduces memory footprint when using disk storage mode by implementing lazy loading of batch content
- **Backward Compatibility**: Fully backward compatible, no changes to public API
- **Testing Coverage**: All existing tests have been updated to work with the new lazy-loading approach

This optimization is particularly beneficial for applications that:
- Process large volumes of events
- Have memory constraints or need to optimize memory usage
- Queue multiple batches for upload simultaneously

The refactor ensures that batch content is only loaded into memory when actually needed for upload, rather than loading all batches during the initial retrieval phase.

### DIFF
--- a/Sources/RudderStackAnalytics/Base/EventManager.swift
+++ b/Sources/RudderStackAnalytics/Base/EventManager.swift
@@ -136,10 +136,14 @@ extension EventManager {
                     // Check shutdown conditions before processing each item
                     if self.analytics.isAnalyticsShutdown { break }
                     
+                    // Read the event batch from storage
+                    let batch = self.analytics.storage.eventStorageMode == .memory ? item.batch : (FileManager.contentsOf(file: item.reference) ?? .empty)
+                    guard !batch.isEmpty else { continue }
+
                     LoggerAnalytics.debug(log: "Upload started: \(item.reference)")
                     do {
                         // Process the batch by replacing timestamp placeholder with current time
-                        let processed = item.batch.replacingOccurrences(of: Constants.payload.sentAtPlaceholder, with: Date().iso8601TimeStamp)
+                        let processed = batch.replacingOccurrences(of: Constants.payload.sentAtPlaceholder, with: Date().iso8601TimeStamp)
                         LoggerAnalytics.debug(log: "Uploading (processed): \(processed)")
                         
                         // Send the batch to the data plane
--- a/Sources/RudderStackAnalytics/Base/EventManager.swift
+++ b/Sources/RudderStackAnalytics/Base/EventManager.swift
@@ -138,7 +138,10 @@ extension EventManager {
                     
                     // Read the event batch from storage
                     let batch = self.analytics.storage.eventStorageMode == .memory ? item.batch : (FileManager.contentsOf(file: item.reference) ?? .empty)
-                    guard !batch.isEmpty else { continue }
+                    guard !batch.isEmpty else { 
+                        LoggerAnalytics.debug(log: "No batch found for reference: \(item.reference)")
+                        continue 
+                    }
 
                     LoggerAnalytics.debug(log: "Upload started: \(item.reference)")
                     do {
--- a/Sources/RudderStackAnalytics/Base/EventManager.swift
+++ b/Sources/RudderStackAnalytics/Base/EventManager.swift
@@ -140,7 +140,10 @@ extension EventManager {
                     let batch = self.analytics.storage.eventStorageMode == .memory ? item.batch : (FileManager.contentsOf(file: item.reference) ?? .empty)
                     guard !batch.isEmpty else { 
                         LoggerAnalytics.debug(log: "No batch found for reference: \(item.reference)")
-                        continue 
+                        
+                        // Remove empty batch from storage
+                        await self.storage.remove(eventReference: item.reference)
+                        continue
                     }
 
                     LoggerAnalytics.debug(log: "Upload started: \(item.reference)")
--- a/Sources/RudderStackAnalytics/Base/EventManager.swift
+++ b/Sources/RudderStackAnalytics/Base/EventManager.swift
@@ -127,8 +127,6 @@ extension EventManager {
             guard let self else { return }
             
             for await _ in self.uploadChannel.receive() {
-                // If shutdown is initiated, don't start new upload cycles
-                if self.analytics.isAnalyticsShutdown { break }
                 
                 // Read all available batched events from storage
                 let dataItems = await self.storage.read().dataItems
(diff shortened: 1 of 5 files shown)


Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
```diff
@@ -184,7 +184,11 @@ extension RudderStackAnalyticsTests {
         try? await Task.sleep(nanoseconds: 300_000_000)
         await client.configuration.storage.rollover()
         let dataItems = await client.configuration.storage.read().dataItems
-        let batchData = dataItems.first?.batch.toDictionary?["batch"] as? [[String: Any]] ?? []
+        
+        guard let firstItem = dataItems.first else { XCTFail("No data item found"); return }
+        let batch = client.storage.eventStorageMode == .memory ? firstItem.batch : (FileManager.contentsOf(file: firstItem.reference) ?? "")
```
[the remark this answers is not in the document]
SKannaniOS: All test cases will be revamped in future.

Tests/RudderStackAnalyticsTests/PluginTests/LifecycleTrackingPluginTests.swift
```diff
@@ -100,7 +100,13 @@ final class LifecycleTrackingPluginTests: XCTestCase {
         await analyticsMock.configuration.storage.rollover()
         
         let dataItems = await analyticsMock.configuration.storage.read().dataItems
-        let batchData = dataItems.first?.batch.toDictionary?["batch"] as? [[String: Any]] ?? []
+        
+        guard let firstDataItem = dataItems.first else {
+            XCTFail("No data items to read"); return []
+        }
+        
+        let batch = analyticsMock.storage.eventStorageMode == .memory ? firstDataItem.batch : (FileManager.contentsOf(file: firstDataItem.reference) ?? "")
```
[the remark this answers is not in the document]
SKannaniOS: All test cases will be revamped in future.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -136,10 +136,14 @@ extension EventManager {
                     // Check shutdown conditions before processing each item
                     if self.analytics.isAnalyticsShutdown { break }
                     
+                    // Read the event batch from storage
+                    let batch = self.analytics.storage.eventStorageMode == .memory ? item.batch : (FileManager.contentsOf(file: item.reference) ?? .empty)
```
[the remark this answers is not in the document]
SKannaniOS: Debug log added.

ChryssaAliferi (approved):
Looks good to me.

Sources/RudderStackAnalytics/Base/EventManager.swift
```diff
@@ -136,10 +136,17 @@ extension EventManager {
...
+                    guard !batch.isEmpty else { 
+                        LoggerAnalytics.debug(log: "No batch found for reference: \(item.reference)")
+                        continue 
+                    }
```
1abhishekpandey: Do you think we should clean up the stale file if no batch is found?

Otherwise, this issue will keep on occurring.

Sources/RudderStackAnalytics/Base/EventManager.swift
1abhishekpandey, line 1: Let's remove [this](https://github.com/rudderlabs/rudder-sdk-swift/blob/86ad16591e3c659d694aedb7cdca8493fc717d85/Sources/RudderStackAnalytics/Base/EventManager.swift#L131) line, I think this is not required, as we are performing the same check again after couple of lines, refer [here](https://github.com/rudderlabs/rudder-sdk-swift/blob/86ad16591e3c659d694aedb7cdca8493fc717d85/Sources/RudderStackAnalytics/Base/EventManager.swift#L137). I think nothing will change within those two checks:

```
if self.analytics.isAnalyticsShutdown { break }

// Read all available batched events from storage
let dataItems = await self.storage.read().dataItems
for item in dataItems {
    // Check shutdown conditions before processing each item
    if self.analytics.isAnalyticsShutdown { break }
```

SKannaniOS: - The first condition is checked before starting the batch upload process.
- The second condition is checked while uploading multiple batches. (Consider when multiple batch items ready to upload)

Therefore, both conditions are required.

1abhishekpandey: Suppose you keep only the second line — would there be any issue? I believe the second check should handle all the edge cases.

Can you please explain this scenario further:
> Consider when multiple batch items ready to upload

I don't see any issue with this, even if we have just the second check.

SKannaniOS: Yes, having second condition only approach covers the scenario, but note that the file is read prior to the shutdown check.

1abhishekpandey: That should be fine. We’re only reading the **available file paths** that need to be flushed, and before flushing we do check whether analytics is shut down. Therefore, I don’t see this as an issue.

It would only be problematic if we tried to flush without performing that second check, or if I had suggested removing the second shutdown check.

To reiterate: reading the file paths (that need to be uploaded) after the shutdown check is acceptable, as long as it doesn’t cause a bug or unexpected behaviour.

SKannaniOS: Ok. I'll check & update.

SKannaniOS: Implementation updated.

1abhishekpandey (approved):

PR #117, opened 2025-08-19
refactor: transform EventManager into EventQueue with separated components

SKannaniOS:
## Description
This PR refactors the event management system by replacing the monolithic `EventManager` class with a modular `EventQueue` architecture. The new design separates concerns into distinct components:
- `EventQueue`: Main coordinator managing the event pipeline
- `EventWriter`: Handles event processing and writing to storage
- `EventUploader`: Manages batch uploading to the data plane

This refactoring improves code maintainability, testability, and follows the single responsibility principle while maintaining the same functionality and API compatibility.

## Type of change
- [ ] Documentation update

## Implementation Details
- **Removed**: `EventManager.swift` - Single class handling all event operations
- **Added**: `EventQueue/EventQueue.swift` - Main coordinator with centralized channel management
- **Added**: `EventQueue/EventWriter.swift` - Dedicated component for event processing and storage operations
- **Added**: `EventQueue/EventUploader.swift` - Dedicated component for uploading events to data plane
- **Updated**: `RudderStackDataPlanePlugin.swift` - Changed from using `EventManager` to `EventQueue`
- **Added**: `EventQueueTests.swift` - Comprehensive test suite for the new architecture
- **Updated**: Xcode project files to reflect new file structure and organization

### Key Architectural Changes:
- **Separation of Concerns**: Split monolithic EventManager into focused components
- **Centralized Communication**: EventQueue manages AsyncChannel instances and passes them to components
- **Improved Lifecycle Management**: Better resource cleanup with explicit component lifecycle
- **Enhanced Testability**: Separated components allow for more granular unit testing
- **Maintained API Compatibility**: All public interfaces remain unchanged

## Checklist

## How to test?
The refactoring includes comprehensive test coverage:

### New Test Cases Added:
1. **Single Event Flow Test**: Verifies that individual events are processed end-to-end from input to storage
2. **Multiple Events Flow Test**: Ensures multiple concurrent events are handled correctly
3. **Stop Functionality Test**: Confirms that stopping the event queue prevents processing of new events

### Test Environment:
- Uses existing `MockProvider.clientWithDiskStorage` for consistent test setup
- Includes proper cleanup and resource management
- Uses async/await patterns for reliable testing of concurrent operations

### Manual Testing:
- Existing example apps continue to work without modification
- All analytics functionality (track, identify, screen, group, alias) remains operational
- Event flushing and upload behavior is preserved
- Memory and performance characteristics are maintained

## Breaking Changes
None. This is a pure refactoring that maintains complete API compatibility. All public interfaces remain unchanged, and existing code using the SDK will continue to work without any modifications.

## Maintainers Checklist
- [ ] CI tests have passed.

N/A - This is an internal architecture refactoring with no UI changes.

## Additional Context
- **Motivation**: The original EventManager class was becoming too large and handling multiple responsibilities, making it harder to maintain and test
- **Benefits**: 
  - Improved code organization and readability
  - Better testability with focused unit tests
  - Easier future maintenance and feature additions
  - Clearer separation of concerns between event processing and uploading
- **Performance**: No performance impact expected as the same underlying logic and async patterns are maintained
- **Future Work**: This architecture enables easier addition of features like retry mechanisms


Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
```diff
@@ -0,0 +1,87 @@
...
+                // Check if we should flush (either explicit flush or policy-triggered)
+                if isFlushSignal || self.flushPolicyFacade.shouldFlush() {
+                    do {
+                        self.flushPolicyFacade.resetCount()
+                        await self.storage.rollover()
+                        
+                        // Only send upload signal if analytics is active
+                        if self.analytics.isAnalyticsActive {
+                            try self.uploadChannel.send(Constants.defaultConfig.uploadSignal)
+                        }
```
1abhishekpandey, line 81: I think this `self.analytics.isAnalyticsActive` check is not needed over here. The EventUploader has the necessary safeguards.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
```diff
@@ -0,0 +1,103 @@
...
+        self.eventWriter = EventWriter(analytics: analytics, writeChannel: self.writeChannel, uploadChannel: self.uploadChannel)
+        self.eventUploader = EventUploader(analytics: analytics, uploadChannel: self.uploadChannel)
+        
+        self.start()
+    }
+    
+    deinit {
+        self.eventWriter = nil
+        self.eventUploader = nil
+    }
```
1abhishekpandey, line 40: Let's not have deinit. Utilise `stop` API.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+}
+
+// MARK: - ProcessingEvent
+/**
+ * ProcessingEvent represents an event in the processing pipeline.
+ * It can be either a regular message event or a flush signal.
+ */
+class ProcessingEvent {
```
1abhishekpandey: Let's move this class into a new file, as now, this class is being used in two files.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
```diff
@@ -0,0 +1,103 @@
...
+        self.type = type
+        self.event = event
+    }
+}
+
+// MARK: - ProcessingEventType
+/**
+ * Enum representing the different types of processing events.
+ */
+enum ProcessingEventType {
```
1abhishekpandey: Let's move this enum into a new file, as now, this enum is being used in two files.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: Why?

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: This enum fits best as `fileprivate`.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: This class has very few lines and it is used inside this module only. Do you really want to create new file?

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
SKannaniOS: This condition prevents control from reaching the upload channel (implemented in another class). As a result, `EventUploader` remains unaffected. Since we are closing upload channel first, we need this condition.

1abhishekpandey: If we are closing the `upload` channel first, isn't the flush request made over here to the upload channel dropped automatically?

I'm trying to ensure, we make a necessary check only if it's needed.

Can you examine what happens when you remove this check and perform a `shutdown` call? Do you get to see any unexpected behaviour?

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: Becuase we already have a `stop` method, which we can utilise for such teardown operations.

Having two different ways for the teardown process might break something in future.

Do you think, if we set those values to `nil` in `stop()`, will it create any issues?

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: Yes, for two reasons:
1. To make the EventQueue class cleaner.
2. For proper visibility of this class.

I would prefer not to add a class within the file, which is being used in more than one file.
As this class is being utilised in two different files. I think it would be better to have a separate file for this class.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: Even though it's `fileprivate`, we are using it over [here](https://github.com/rudderlabs/rudder-sdk-swift/blob/f762c97c2828435d142a7b5773982d6fffa72e21/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift#L59) and [here](https://github.com/rudderlabs/rudder-sdk-swift/blob/f762c97c2828435d142a7b5773982d6fffa72e21/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift#L35) in the `EventWriter` class.

Since we are sharing the ENUM in two different classes, I think we should try to define these enums in a separate file.

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
SKannaniOS: If we remove the mentioned check & sends the flush signal means, the `uploadChannel` would throw an `error`, which we already handled with a `do-catch` block. However, this condition ensures that the upload channel is never reached during a shutdown, which is the good, I believe.

1abhishekpandey: If this is the case:
> If we remove the mentioned check & sends the flush signal means, the uploadChannel would throw an error

Then it's perfectly fine to have this check. I would resolve this comment.

Just sharing my thoughts on this:
>  However, this condition ensures that the upload channel is never reached during a shutdown, which is the good, I believe.

You're right, but I think we should try to avoid unnecessary checks to make the flow simple and easy to read. Since, uploadChannel is already having a check, ideally, I expect it to handle all the edge cases, like our Kotlin SDK does. It makes the code easy to read.
Ideally, shutdown operation should only affect the upload process and not the storage process. So, we should have a check only at the upload class. This is my expectation and this is what we have implemented successfully in the Kotlin. Refer to the Kotlin code [here](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/3950978ed60f8e3322232571dcfd3396aad5265d/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/queue/EventQueue.kt#L117-L120), for reference.

I'm resolving this comment, but I would request you to explore if we can skip this check by modifying anything. 

I repeat again, if it's necessary and can't be avoided (due to the exception reason you mentioned) then it's perfectly fine to have this check, as we know **the main purpose of this check is to avoid the exception**.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: In general, we don’t need to explicitly assign nil to optional properties. However, if you want to ensure that certain properties are released when a class instance is deallocated, `deinit()` is the right place. If you’re not comfortable with having a `deinit()` implementation, I can remove it — it won’t cause any issues.

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
SKannaniOS: Usually I don’t add these kinds of checks, but consider them more as fail-safes. Anyway, I’m aligned with your opinion.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: Done.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: I moved this `enum` into the `ProcessingEvent` class since it scoped to that class improves readability and encapsulation.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: I got you.

Regarding this:
> If you’re not comfortable with having a deinit() implementation, I can remove it — it won’t cause any issues.

Let me explain my thought process:
- Earlier, I faced some issues due to the `deinit()` (I cannot recall the reason; I need to find if required) and also we have a `stop()` method; due to these two reasons, I recommended updating the implementation.
- Given, we already have a `stop()` / `teardown()`, I thought we should prefer that over `deinit()`.
- Also, I can see we don't actively use `deinit()` in other plugin classes (except a few). **This seems confusing, as it's not clear to me, when we are trying to prefer `deinit()` vs. `teardown()`.**
- As per my understanding, we should consider `deinit()` only on those classes which is not of plugin type and are independent i.e., any plugin classes do not control them. Because if it's a plugin class or controlled by a plugin class, then we will have the `teardown()`.

Also, I would recommend that you consider removing [this](https://github.com/rudderlabs/rudder-sdk-swift/blob/30de0d819328b8991d92a7be38d74dca9e2702b2/Sources/RudderStackAnalytics/Plugins/RudderStackDataPlanePlugin.swift#L25-L28) and use the plugin `teardown()`.

Having consistency in the implementation in this case would be better, i.e., prefer `teardown()` for all plugin classes and then `deinit()` otherwise.

If you still think `deinit()` would be helpful and necessary, then please proceed ahead with this. Just reply to this comment and I'll mark it as resolved.

1abhishekpandey: @SKannaniOS Let's prepare a plan for when to prefer `deinit()` and when to prefer plugin `teardown()`. We should try to have consistency in the implementation.

We can follow this plan going forward.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: Great.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
SKannaniOS: The purpose of `deinit()` is to perform actions just before an instance is released. Since it’s automatically invoked, we can use `deinit()` whenever we need an automated `exit-time` execution in `Swift` implementations. Going forward, my proposal is to adopt `deinit()` for such cases.

1abhishekpandey: This is a bit unclear to me:
> Going forward, my proposal is to adopt deinit() for such cases.

Which exact cases are we referring to? Can you please elaborate on this?

Also, what's your proposal on the use of `teardown()`? Also, suggest when to use `deinit()` and `teardown()`. Or, do you propose to completely remove the `teardown()`?

SKannaniOS: We can adopt `deinit()` when execution is acceptable at the time of instance release. If the execution needs to happen immediately, we should use an explicit `teardown()` method.

**For example:**
	•	If we need to change the state of a property and the change must take effect immediately, we should prefer `teardown()`.
	•	If we only need to release resources or properties when the instance is deallocated, then `deinit()` is the right choice.

In `Swift`, both approaches are valid—the decision fully depends on the specific need or scenario. If `Kotlin` had a `deinit-like` mechanism (an automatic cleanup hook on object destruction), I’m sure you would prefer it over using `teardown()`.

Cheers.. 👍

Sources/RudderStackAnalytics/EventQueue/ProcessingEvent.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  ProcessingEvent.swift
+//  RudderStackAnalyticsTests
```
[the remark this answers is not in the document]
SKannaniOS: Done.

Tests/RudderStackAnalyticsTests/EventQueueTests.swift
```diff
@@ -0,0 +1,170 @@
...
+                    ? item.batch
+                    : (FileManager.contentsOf(file: item.reference) ?? .empty)
+                    return batch.contains("integration_test")
+                }) {
+                    expectation.fulfill()
+                    break
+                }
+                
+                await Task.yield()
+            }
```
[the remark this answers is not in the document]
SKannaniOS: Leaving it for now — will be covered during the test case revamp.

Tests/RudderStackAnalyticsTests/EventQueueTests.swift
```diff
@@ -0,0 +1,170 @@
...
+                    if batch.contains("event_3") { found.insert("event_3") }
+                }
+                
+                if found.count == events.count {
+                    expectation.fulfill()
+                    break
+                }
+                
+                await Task.yield()
+            }
```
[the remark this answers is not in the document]
SKannaniOS: Leaving it for now — will be covered during the test case revamp.

Tests/RudderStackAnalyticsTests/EventQueueTests.swift
```diff
@@ -0,0 +1,170 @@
...
+            expectation.fulfill()
+        }
+        
+        await fulfillment(of: [expectation], timeout: 2.0)
+        
+        // Cleanup any existing events
+        let dataItems = await mockAnalytics.configuration.storage.read().dataItems
+        for item in dataItems {
+            await mockAnalytics.configuration.storage.remove(eventReference: item.reference)
+        }
```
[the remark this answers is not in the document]
SKannaniOS: Leaving it for now — will be covered during the test case revamp.

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: Regarding this:
> We can adopt deinit() when execution is acceptable at the time of instance release. If the execution needs to happen immediately, we should use an explicit teardown() method.

and this:
> For example:
If we only need to release resources or properties when the instance is deallocated, then deinit() is the right choice.

Do you see any issue if we perform all the actions in the `teardown()` method? Will it have any impact?

I see an advantage, with having only one method, i.e., if we try to stick to only one method, i.e., either `teardown` or `deinit`, it'll improve the code maintainibility, we don't need to perform same cleanup operation in two different methods.

Regarding this:
> If Kotlin had a deinit-like mechanism (an automatic cleanup hook on object destruction), I’m sure you would prefer it over using teardown().

Even if Kotlin had a `deinit()` like mechanism and given our project already provides a `teardown()` method, I’d prefer sticking to just one of them. Using two different approaches for the same cleanup task doesn’t seem beneficial to me.

Let's discuss this offline.

ChryssaAliferi (approved):

Sources/RudderStackAnalytics/EventQueue/EventQueue.swift
1abhishekpandey: We had a discussion and aligned on the following points:
- We’ll have a `teardown()` method in the Swift SDK plugin interface and it’ll be an optional method for users to override, in case they want to use it in the CustomPlugins.
    - Apart from this `teardown()` doesn’t have any other use-case inside the SDK.
- Swift SDK will not use this `teardown()` method anywhere inside the SDK.
- Instead, Swift SDK will use `deinit()` method wherever necessary.

To summarise, Swift SDK will use `deinit()` in place of `teardown()`.

CC: @SKannaniOS @ChryssaAliferi

1abhishekpandey (approved):

PR #122, opened 2025-08-26
feat: implement storage api to remove batch files by write key

SKannaniOS:
## Description
This PR implements a new storage API that enhances batch file management with write key-based isolation and bulk removal capabilities. The changes introduce a `removeAll` method for efficient cleanup of all batches associated with a specific write key, along with improved file organization and better semantic clarity by renaming methods from event-based to batch-based terminology.

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Implementation Details
- **Storage Protocol Enhancement**: Added `removeAll(batchReference:)` method to `EventStorage` protocol for bulk batch removal operations
- **API Method Renaming**: Renamed `remove(eventReference:)` to `remove(batchReference:)` across all storage implementations for better semantic clarity
- **Write Key Directory Structure**: Modified `DiskStore` to organize files under write key-specific subdirectories (`{base_path}/{writeKey}/`)
- **File Management Improvements**: 
  - Added `createDirectoryIfNeeded(at:)` helper method for safe directory creation
  - Enhanced `delete(item:)` method to handle both files and directories
  - Simplified file naming convention by removing write key from filenames (now index-based only)
- **Memory Store Enhancements**:
  - Implemented write key prefixing for batch references to ensure data isolation
  - Added helper methods `appendWriteKey(with:)` and `recoverWriteKey(from:)` for reference management
  - Added `removeAll(reference:)` implementation for memory-based bulk removal
- **Constants Addition**: Added `referenceSeparator` constant for consistent write key reference formatting
- **Comprehensive Test Updates**: Updated all test files to use the new `remove(batchReference:)` API and added test coverage for `removeAll` functionality

## Checklist

## How to test?
1. **Disk Storage Testing**: 
   - Verify that files are created in write key-specific directories
   - Test individual batch removal using `remove(batchReference:)`
   - Test bulk removal using `removeAll(batchReference:)` 
   - Confirm proper directory cleanup after bulk removal

2. **Memory Storage Testing**:
   - Verify write key prefixing in batch references
   - Test isolation between different write keys
   - Test bulk removal functionality for memory storage

## Breaking Changes
- **Method Signature Change**: `remove(eventReference: String)` has been renamed to `remove(batchReference: String)`
- **Directory Structure Change**: Files are now organized under write key-specific subdirectories instead of the shared directory

## Maintainers Checklist
- [ ] CI tests have passed.

N/A - This is a backend storage API enhancement without UI changes.

## Additional Context
- **Performance Impact**: The new directory structure and write key isolation provide better organization and should improve performance for applications using multiple write keys
- **Security Enhancement**: Write key isolation prevents cross-contamination of data between different analytics instances
- **Future Compatibility**: This change sets the foundation for better multi-tenant support and more efficient batch management

### File Organization Changes:
**Before**: `rudder/analytics/events/{writeKey}-{index}.tmp`  
**After**: `rudder/analytics/events/{writeKey}/{index}.tmp`

### API Changes:
```swift
// Old API
func remove(eventReference: String) async -> Bool

// New API  
func remove(batchReference: String) async -> Bool
func removeAll(batchReference: String) async
```

### DIFF
--- a/Sources/RudderStackAnalytics/Storage/BasicStorage.swift
+++ b/Sources/RudderStackAnalytics/Storage/BasicStorage.swift
@@ -50,8 +50,8 @@ extension BasicStorage {
         return await self.dataStore.remove(reference: batchReference)
     }
     
-    func removeAll(batchReference: String) async {
-        await self.dataStore.removeAll(reference: batchReference)
+    func removeAll() async {
+        await self.dataStore.removeAll()
     }
     
     func rollover() async {
--- a/Sources/RudderStackAnalytics/Storage/BasicStorage.swift
+++ b/Sources/RudderStackAnalytics/Storage/BasicStorage.swift
@@ -50,6 +50,10 @@ extension BasicStorage {
         return await self.dataStore.remove(reference: batchReference)
     }
     
+    func removeAll(batchReference: String) async {
+        await self.dataStore.removeAll(reference: batchReference)
+    }
+    
     func rollover() async {
         await self.dataStore.rollover()
     }
--- a/Sources/RudderStackAnalytics/Storage/BasicStorage.swift
+++ b/Sources/RudderStackAnalytics/Storage/BasicStorage.swift
@@ -46,8 +46,8 @@ extension BasicStorage {
         return await EventDataResult(dataItems: self.dataStore.retrieve())
     }
     
-    func remove(eventReference: String) async -> Bool {
-        return await self.dataStore.remove(reference: eventReference)
+    func remove(batchReference: String) async -> Bool {
+        return await self.dataStore.remove(reference: batchReference)
     }
     
     func rollover() async {
--- a/Sources/RudderStackAnalytics/Storage/Storage.swift
+++ b/Sources/RudderStackAnalytics/Storage/Storage.swift
@@ -82,11 +82,9 @@ protocol EventStorage {
     func remove(batchReference: String) async -> Bool
 
     /**
-     Removes all batches associated with the given reference from the storage.
-
-     - Parameter batchReference: The reference of the batches to be removed.
+     Removes all batches associated with the current write key.
      */
-    func removeAll(batchReference: String) async
+    func removeAll() async
     
     /**
      Performs a rollover operation on the storage.
--- a/Sources/RudderStackAnalytics/Storage/Storage.swift
+++ b/Sources/RudderStackAnalytics/Storage/Storage.swift
@@ -73,14 +73,14 @@ protocol EventStorage {
     func read() async -> EventDataResult
 
     /**
-     Removes a specific event from the storage.
+     Removes a specific batch from the storage.
 
-     - Parameter eventReference: The reference of the event to be removed.
-     - Returns: A `Bool` indicating whether the event was successfully removed.
+     - Parameter batchReference: The reference of the batch to be removed.
+     - Returns: A `Bool` indicating whether the batch was successfully removed.
      */
     @discardableResult
-    func remove(eventReference: String) async -> Bool
-
+    func remove(batchReference: String) async -> Bool
+    
     /**
      Performs a rollover operation on the storage.
 
--- a/Sources/RudderStackAnalytics/Storage/Storage.swift
+++ b/Sources/RudderStackAnalytics/Storage/Storage.swift
@@ -80,6 +80,13 @@ protocol EventStorage {
      */
     @discardableResult
     func remove(batchReference: String) async -> Bool
+
+    /**
+     Removes all batches associated with the given reference from the storage.
+
+     - Parameter batchReference: The reference of the batches to be removed.
+     */
+    func removeAll(batchReference: String) async
     
     /**
      Performs a rollover operation on the storage.
--- a/Sources/RudderStackAnalytics/Storage/Storage.swift
+++ b/Sources/RudderStackAnalytics/Storage/Storage.swift
@@ -83,6 +83,8 @@ protocol EventStorage {
 
     /**
      Removes all batches associated with the current write key.
+
+     Note: This operation is irreversible and will permanently delete all associated events.
      */
     func removeAll() async
     
--- a/Sources/RudderStackAnalytics/Storage/Storage.swift
+++ b/Sources/RudderStackAnalytics/Storage/Storage.swift
@@ -84,7 +84,7 @@ protocol EventStorage {
     /**
      Removes all batches associated with the current write key.
 
-     Note: This operation is irreversible and will permanently delete all associated events.
+     **Note**: It is recommended to use this API during shutdown to ensure storage is not removed abruptly, which could lead to unexpected errors.
      */
     func removeAll() async
     
--- a/Sources/RudderStackAnalytics/Storage/Store/DiskStore.swift
+++ b/Sources/RudderStackAnalytics/Storage/Store/DiskStore.swift
@@ -61,7 +61,11 @@ final actor DiskStore {
         return FileManager.contentsOf(directory: directory.path)
             .filter { $0.pathExtension.isEmpty }
             .map { directory.appendingPathComponent($0.lastPathComponent).path }
-            .sorted { (Int($0) ?? 0) < (Int($1) ?? 0) }
+            .sorted {
+                let lhsIndex = Int(URL(fileURLWithPath: $0).lastPathComponent) ?? 0
+                let rhsIndex = Int(URL(fileURLWithPath: $1).lastPathComponent) ?? 0
+                return lhsIndex < rhsIndex
+            }
     }
     
     private func removeItems(using reference: String) {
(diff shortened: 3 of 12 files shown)


Sources/RudderStackAnalytics/Storage/Store/DiskStore.swift
```diff
@@ -54,14 +59,20 @@ final actor DiskStore {
...
+                let lhsIndex = Int(URL(fileURLWithPath: $0).lastPathComponent) ?? 0
+                let rhsIndex = Int(URL(fileURLWithPath: $1).lastPathComponent) ?? 0
+                return lhsIndex < rhsIndex
             }
     }
+    
+    private func removeItems(using reference: String) {
+        let folderPath = (reference as NSString).deletingLastPathComponent
+        let result = FileManager.delete(item: folderPath)
+        result ? LoggerAnalytics.debug(log: "Successfully removed folder: \(folderPath)") : LoggerAnalytics.debug(log: "Folder does not exist: \(folderPath)")
```
1abhishekpandey: Let's avoid printing duplicate log messages. The first message is printed over [here](https://github.com/rudderlabs/rudder-sdk-swift/blob/66bdab3864de834fc35c97a50261eef6b0061aa5/Sources/RudderStackAnalytics/Helpers/Extensions.swift#L131), just after the delete operation and then again over here.

Here's the log:
```
[Rudder-Analytics] DEBUG: Item deleted: /Users/[USERNAME]/Library/Developer/CoreSimulator/Devices/[DEVICE_ID]/data/Containers/Data/Application/[APP_ID]/Documents/rudder/analytics/events/[EVENT_ID]
[Rudder-Analytics] DEBUG: Successfully removed folder: /Users/[USERNAME]/Library/Developer/CoreSimulator/Devices/[DEVICE_ID]/data/Containers/Data/Application/[APP_ID]/Documents/rudder/analytics/events/[EVENT_ID]
```

I think we can optimise it further by printing the path at only one place.

1abhishekpandey: As discussed offline, let's remove this log statement from here and add the log statement directly in the `delete` function `catch` block.

```suggestion
```

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
```diff
@@ -46,8 +46,12 @@ extension BasicStorage {
...
     }
     
-    func remove(eventReference: String) async -> Bool {
-        return await self.dataStore.remove(reference: eventReference)
+    func remove(batchReference: String) async -> Bool {
+        return await self.dataStore.remove(reference: batchReference)
+    }
+    
+    func removeAll(batchReference: String) async {
+        await self.dataStore.removeAll(reference: batchReference)
```
1abhishekpandey: I think there's a possibility to optimise this step further. We don't really need this `batchReference` parameter, as we are targeting to delete the whole directory. We can find the current directory path with the help of `writeKey`.

Proposal:
- Remove the `batchReference` parameter.
- Construct the directory path somewhere else.

Note: We've implemented the same approach in Android.

1abhishekpandey: Also, I would request to rename the `removeAll` to something more meaningful that denotes we are removing the whole directory.

You can have it like `deleteDirectory`, or anything of your choice.

Sources/RudderStackAnalytics/Storage/Storage.swift
```diff
@@ -73,14 +73,21 @@ protocol EventStorage {
...
      */
     @discardableResult
-    func remove(eventReference: String) async -> Bool
+    func remove(batchReference: String) async -> Bool
 
+    /**
+     Removes all batches associated with the given reference from the storage.
+
+     - Parameter batchReference: The reference of the batches to be removed.
+     */
```
1abhishekpandey: Let's update the description to highlight two things:
1. State clearly that we are targeting to delete the whole directory itself.
2. Add a note to use this API with caution.

Refer to the [Kotlin doc of this API](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/5963e772b9fff115350679cabba6ec7089fd8c03/core/src/main/kotlin/com/rudderstack/sdk/kotlin/core/internals/storage/Storage.kt#L138-L146).

1abhishekpandey: Note: I would also encourage you to add another note in the doc to indicate that this should be called as part of the shutdown operation.

Reason:
- If someone delete the storage completely and then suppose SDK tries to perform some event storage operation. SDK will crash!

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
```diff
@@ -46,8 +46,12 @@ extension BasicStorage {
         return await EventDataResult(dataItems: self.dataStore.retrieve())
     }
     
-    func remove(eventReference: String) async -> Bool {
-        return await self.dataStore.remove(reference: eventReference)
+    func remove(batchReference: String) async -> Bool {
+        return await self.dataStore.remove(reference: batchReference)
+    }
+    
+    func removeAll(batchReference: String) async {
```
1abhishekpandey: You're deleting the event storage directory successfully, but I think `UserDefault` is never cleared. 

Can you think more about it? On how to clear UserDefault? And I would recommend that you create an incremental PR to implement this change.

1abhishekpandey: If you've already planned to do that in another PR, then you can ignore it and just provide the linear ticket number to me.

1abhishekpandey (commented):
Great work.

Sources/RudderStackAnalytics/Storage/Storage.swift
SKannaniOS: This method is invoked only after shutdown is performed. Not part of the public API.

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
SKannaniOS: What are the details you want clear from `UserDefaults`?

1abhishekpandey: In short, the UserDefault file itself for that specific writeKey.

Whatever value is stored in the UserDefault for the particular `writeKey` should be removed, as part of this removeAll operation.

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
1abhishekpandey: Are we planning to rename the method? I think that comment is yet to be addressed!

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
SKannaniOS: Is that includes all user details? (like anonumoysId, userId & traits)

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
SKannaniOS: Method name `removeAll` is self-explanatory in the context of `Storage`. Since a method for removing a single item already exists, this one clearly indicates removal of all items.

Note: This is not a public method.

Sources/RudderStackAnalytics/Storage/Storage.swift
1abhishekpandey: I proposed to add a disclaimer (or a kind of warning) by documenting the fact that this API should be called only as part of the shutdown operation. 

Motive: This is to ensure that future contributors who are planning to use this API will have sufficient guidelines on how to use this internal API correctly.

As of now, this is the API doc:
> /**
     Removes all batches associated with the current write key.

     Note: This operation is irreversible and will permanently delete all associated events.
     */
    func removeAll() async

I'm proposing to have it like this:
```
/**
  Removes all batches associated with the current write key.

  **Note**: It is recommended to use this API during shutdown to ensure storage is not removed abruptly, which could lead to unexpected errors.
  
  Note: This operation is irreversible and will permanently delete all associated events.
*/
func removeAll() async
```

Sources/RudderStackAnalytics/Storage/BasicStorage.swift
1abhishekpandey: Yes. Everything that is stored by our SDK for that specific writeKey in the UserDefault, including the value you mentioned and also other values like SourceConfig, sessionId ... etc. and other future values.

We should have a way to delete this file, at once.

SKannaniOS: But we are not doing the same as a part of Shutdown API. Right?

1abhishekpandey: Let's have an internal discussion on this.

SKannaniOS: As per the internal discussed, I'll address this change in the separate PR.

1abhishekpandey (approved):

PR #137, opened 2025-09-15
feat: add generic plugin lookup to ensure proper configuration before initialization

anahjz:
## Description
This PR introduces a generic find<T: Plugin> helper method that allows safe retrieval of plugins by type from the plugin chain. This enhancement was developed to solve configuration issues with PushNotificationPlugin where the plugin was already initialized during Analytics setup, causing subsequent calls to configurePushNotifications to fail to properly update the plugin's configuration.

## Type of change

- [ ] Documentation update

## Implementation Details
Added generic find<T: Plugin>(type: T.Type) -> T? method to Analytics class
Enables type-safe plugin retrieval from the plugin chain
Allows runtime plugin configuration updates after Analytics initialization
Solves plugin reconfiguration issues for push notifications and other plugins

## Checklist

## How to test?
Initialize Analytics with default configuration
Use the new find method to retrieve a specific plugin:
`   let analytics = Analytics(configuration: config)
   if let pushPlugin = analytics.find(type: PushNotificationPlugin.self) {
       // Update plugin configuration
       pushPlugin.updateConfiguration(newConfig)
   }`
Verify that plugin configuration updates work correctly after initialization
Test with different plugin types to ensure generic functionality

## Breaking Changes
None - this is a purely additive feature that maintains backward compatibility.

## Maintainers Checklist

- [ ] CI tests have passed.

## Summary by CodeRabbit

- **New Features**
  - Added a public API to locate and retrieve a plugin instance by type from the analytics client; searches across the plugin chain and returns nil if not found or analytics is inactive.

- **Tests**
  - Added unit tests verifying plugin lookup returns the correct instance and nil after removal.
  - Added identify-reset behavior tests covering various userId/anonymousId scenarios for disk-backed analytics.

### DIFF
--- a/Sources/RudderStackAnalytics/Base/Analytics.swift
+++ b/Sources/RudderStackAnalytics/Base/Analytics.swift
@@ -250,7 +250,8 @@ extension Analytics {
      - Parameter type: The type of plugin to find
      - Returns: The plugin instance if found, nil otherwise
      */
-    public func find<T: Plugin>(type: T.Type) -> T? {
+    public func find<T: Plugin>(_ type: T.Type) -> T? {
+        guard self.isAnalyticsActive else { return nil}
         return pluginChain?.find(type: type)
     }
     
--- a/Sources/RudderStackAnalytics/Base/Analytics.swift
+++ b/Sources/RudderStackAnalytics/Base/Analytics.swift
@@ -243,6 +243,17 @@ extension Analytics {
         guard self.isAnalyticsActive else { return }
         self.pluginChain?.remove(plugin: plugin)
     }
+    
+    /**
+     Finds a plugin of the specified type in the plugin chain.
+
+     - Parameter type: The type of plugin to find
+     - Returns: The plugin instance if found, nil otherwise
+     */
+    public func find<T: Plugin>(type: T.Type) -> T? {
+        return pluginChain?.find(type: type)
+    }
+    
 }
 
 // MARK: - Shutdown
--- a/Sources/RudderStackAnalytics/Plugins/Base/PluginChain.swift
+++ b/Sources/RudderStackAnalytics/Plugins/Base/PluginChain.swift
@@ -21,7 +21,7 @@ class PluginChain {
         PluginType.allCases.forEach { self.pluginList[$0] = PluginInteractor() }
     }
     
-    func process(event: Event) {        
+    func process(event: Event) {
         let preProcessedResult = self.applyPlugins(pluginType: .preProcess, event: event)
         let onProcessedResult = self.applyPlugins(pluginType: .onProcess, event: preProcessedResult)
         
@@ -52,10 +52,11 @@ class PluginChain {
             mediator.removeAll()
         }
     }
-
+    
     func find<T: Plugin>(type: T.Type) -> T? {
-        for (_, mediator) in pluginList {
-            if let found = mediator.find(type) {
+        for pluginType in PluginType.allCases {
+            if let mediator = pluginList[pluginType],
+               let found = mediator.find(type) {
                 return found
             }
         }
--- a/Sources/RudderStackAnalytics/Plugins/Base/PluginChain.swift
+++ b/Sources/RudderStackAnalytics/Plugins/Base/PluginChain.swift
@@ -52,6 +52,15 @@ class PluginChain {
             mediator.removeAll()
         }
     }
+
+    func find<T: Plugin>(type: T.Type) -> T? {
+        for (_, mediator) in pluginList {
+            if let found = mediator.find(type) {
+                return found
+            }
+        }
+        return nil
+    }
 }
 
 // MARK: - Private functions
--- a/Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
+++ b/Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
@@ -259,6 +259,22 @@ extension RudderStackAnalyticsTests {
             await client.configuration.storage.remove(batchReference: item.reference)
         }
     }
+
+    func test_findPlugin_disk() {
+        guard let client = analytics_disk else { return XCTFail("No disk client") }
+        
+        let customPlugin = MockPlugin()
+        client.add(plugin: customPlugin)
+        
+        let foundPlugin = client.find(type: MockPlugin.self)
+        XCTAssertNotNil(foundPlugin, "Should find the added plugin")
+        XCTAssertTrue(foundPlugin === customPlugin, "Should return the same plugin instance")
+        
+        client.remove(plugin: customPlugin)
+        
+        let removedPlugin = client.find(type: MockPlugin.self)
+        XCTAssertNil(removedPlugin, "Should not find plugin after removal")
+    }
 }
 
 // MARK: - Identify Reset Behavior Tests
--- a/Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
+++ b/Tests/RudderStackAnalyticsTests/RudderStackAnalyticsTests.swift
@@ -266,13 +266,13 @@ extension RudderStackAnalyticsTests {
         let customPlugin = MockPlugin()
         client.add(plugin: customPlugin)
         
-        let foundPlugin = client.find(type: MockPlugin.self)
+        let foundPlugin = client.find(MockPlugin.self)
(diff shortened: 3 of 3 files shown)


SKannaniOS:
@anahjz Thanks for your contribution! We'll review this PR shortly.

SKannaniOS:
@anahjz After reviewing the `How to test?` section, if a user holds a reference to the custom plugin and updates its values or configuration at any time, those changes will automatically be reflected in the next event payload. Could you provide more context on why this additional API is strongly needed?

anahjz:
> @anahjz After reviewing the `How to test?` section, if a user holds a reference to the custom plugin and updates its values or configuration at any time, those changes will automatically be reflected in the next event payload. Could you provide more context on why this additional API is strongly needed?


 users can't hold references to plugins that are auto-added internally during initialization. The find method is the only way to access these plugin instances after they're created, the find method is literally the only public API to access these internally-created plugin instances.

SKannaniOS:
@anahjz 
I might be missing something here—could you help me understand how the plugins are added automatically without any reference? My assumption was that users could hold their plugin references and update them directly. If that’s not the case, I’d love a bit more detail on why the find method is the only option.

anahjz:
> @anahjz I might be missing something here—could you help me understand how the plugins are added automatically without any reference? My assumption was that users could hold their plugin references and update them directly. If that’s not the case, I’d love a bit more detail on why the find method is the only option.

In RudderStack Swift, not all plugins are created by the user. Some plugins are added automatically by the SDK during initialization. I used PushPlugin as a automatic plugin.

For example:

Utility plugins like ContextPlugin, SessionPlugin are inserted into the plugin chain internally by Analytics during setup.

Since the SDK manages their lifecycle, users don’t have the opportunity to instantiate or keep references to these plugins directly.

Because of that:

If a user wants to update configuration or runtime values for one of these internal plugins after initialization, there’s currently no supported way to do so.

The find API solves this gap, it lets the user safely access an already initialized plugin instance of a given type and update it (like  updating push notification config)

Without this API, users would have to rely on undocumented assumptions about plugin internals or duplicate the plugin, which risks breaking the pipeline.

So the find API is not about replacing the ability to hold references when you add your own custom plugin, it’s specifically about giving users access to plugins that are created automatically by RudderStack and otherwise inaccessible.

SKannaniOS:
We appreciate your explanation and would like to clarify our understanding. We’re curious why the user wants to update the internal plugins, since updates to the payload or configuration can already be handled through custom plugins.

In the example shared, `PushPlugin` (which adds the push notification token as `context.device.token`) is actually an external plugin — our SDK does not provide an internal one for this purpose. If it’s being used as an external plugin, the user can manage and update its values externally, and the changes will be reflected immediately in subsequent events.

If possible, could you share a snippet showing how the plugin is being initialized and handled? That would help us provide more specific guidance.

anahjz:
> this

In our case, the PushNotificationPlugin is not explicitly created by the user as a custom plugin. Instead, it’s auto-added internally during Analytics initialization when configurePushNotifications() is called. Because the SDK itself creates and manages that plugin, the user has no direct reference to it.
For example, a user may want to update notification settings after app launch:
`
self.analytics?.configurePushNotifications(PushNotificationConfig(
            showInAppNotifications: true,
            doNotShowAlertForNotifications: false
 ))
`
Without the find API, the only alternative would be to create another plugin instance manually, which leads to duplication and inconsistent state.

1abhishekpandey:
Hey @anahjz,

Could you please clarify your current setup, specifically:
1. How you’ve integrated our Swift SDK into your app?
2. How is the `PushNotificationPlugin` being added, given that our SDK doesn’t include it as part of the standard set?

Regarding this:
> In our case, the PushNotificationPlugin is not explicitly created by the user as a custom plugin.

Please clarify whom you are referring to when you say “the user”. (From my understanding, it should be you who adds that custom plugin.)

Regarding this:
> Instead, it’s auto-added internally during Analytics initialization when configurePushNotifications() is called.

Could you please explain how it’s being auto-added? Is the code responsible for this managed by you or by us? Any additional details would be helpful.

You have shared this code snippet:
> self.analytics?.configurePushNotifications(PushNotificationConfig( showInAppNotifications: true, doNotShowAlertForNotifications: false ))

But we don’t provide any API called `configurePushNotifications`. Could you please share more details about this API, such as where it’s defined? Share the plugin code as well.

We’re trying to understand your use case better so we can assist you effectively.

anahjz:
> Hey @anahjz,
> 
> Could you please clarify your current setup, specifically:
> 
> 1. How you’ve integrated our Swift SDK into your app?
> 2. How is the `PushNotificationPlugin` being added, given that our SDK doesn’t include it as part of the standard set?
> 
> Regarding this:
> 
> > In our case, the PushNotificationPlugin is not explicitly created by the user as a custom plugin.
> 
> Please clarify whom you are referring to when you say “the user”. (From my understanding, it should be you who adds that custom plugin.)
> 
> Regarding this:
> 
> > Instead, it’s auto-added internally during Analytics initialization when configurePushNotifications() is called.
> 
> Could you please explain how it’s being auto-added? Is the code responsible for this managed by you or by us? Any additional details would be helpful.
> 
> You have shared this code snippet:
> 
> > self.analytics?.configurePushNotifications(PushNotificationConfig( showInAppNotifications: true, doNotShowAlertForNotifications: false ))
> 
> But we don’t provide any API called `configurePushNotifications`. Could you please share more details about this API, such as where it’s defined? Share the plugin code as well.
> 
> We’re trying to understand your use case better so we can assist you effectively.

let me clarify.
I’m not currently integrating your Swift SDK inside a production app. My goal here is to explore what the integration flow would look like if PushNotificationPlugin were to become an official plugin or a built-in feature of your SDK. I’m doing this to better understand what the ideal integration experience would be for SDK users.

Specifically, I was experimenting with the idea that:

The SDK could provide a Push plugin that handles token registration and displaying notifications.

This plugin would be part of the SDK’s features, not something the end-user has to write manually.

During the integration process, a user of the SDK would just enable it via configuration (e.g., configurePushNotifications(...) or similar) rather than creating a custom plugin themselves.

So when I said “auto added,” I meant conceptually: if your SDK shipped with Push as an optional feature, calling a config method could internally add the plugin to the chain — without requiring the user to write plugin boilerplate themselves.

In other words, I’m not asking about existing APIs (since I know configurePushNotifications doesn’t exist in your SDK). I wrote that method as a hypothetical example to test how the flow might look if Push were first-class in your SDK.

My main question is: if you were to ship Push as an official plugin/feature, would you envision it working this way (via configuration), or as a separate plugin that the user explicitly imports and adds themselves?

1abhishekpandey:
Thanks for clarifying it. 

I understand—you’re trying to explore the PushNotificationPlugin use case.

Regarding this:
> My main question is: if you were to ship Push as an official plugin/feature, would you envision it working this way (via configuration), or as a separate plugin that the user explicitly imports and adds themselves?

Since this is a new feature we haven’t built yet, we can’t confidently say how it will be supported. However, if we do include it in the standard plugin set, we would provide the necessary APIs or SDK configurations to pass the required details—without relying on the `findPlugin` API.

I hope this clarifies the details. Please let me know if you have any further questions.

anahjz:
> Thanks for clarifying it.
> 
> I understand—you’re trying to explore the PushNotificationPlugin use case.
> 
> Regarding this:
> 
> > My main question is: if you were to ship Push as an official plugin/feature, would you envision it working this way (via configuration), or as a separate plugin that the user explicitly imports and adds themselves?
> 
> Since this is a new feature we haven’t built yet, we can’t confidently say how it will be supported. However, if we do include it in the standard plugin set, we would provide the necessary APIs or SDK configurations to pass the required details—without relying on the `findPlugin` API.
> 
> I hope this clarifies the details. Please let me know if you have any further questions.

Thanks for clarifying , that makes sense.

I understand this feature hasn’t been built yet, and that the exact integration path will depend on your design decisions. My main goal here was just to validate whether the concept of Push as a standard plugin (enabled via configuration rather than custom code) aligns with how you see potential future extensions of the SDK.

Your point about providing dedicated APIs or SDK configurations instead of relying on findPlugin is clear, and it answers what I needed.

I don’t have further questions at this point I appreciate your input!

1abhishekpandey:
Shall we go ahead and close this PR?

anahjz:
> Shall we go ahead and close this PR?

Yes, everything looks clear now. We can go ahead and close this PR. Thanks for the discussion and clarifications!

PR #144, opened 2025-09-19
feat: implement event batching based on anonymousId with automatic rollover

1abhishekpandey:
## Description

This PR implements intelligent event batching logic that automatically creates separate batches when the anonymousId changes between events. This ensures that events from different anonymous users are properly isolated in their own batches.

## Type of change

- [ ] Documentation update

### Event Batching Logic
- **EventWriter Enhancement**: Added `lastEventAnonymousId` property to track the anonymousId of the most recent processed event
- **Automatic Rollover**: Implemented `updateAnonymousIdAndRolloverIfNeeded()` method that triggers storage rollover when anonymousId changes between consecutive events
- **Persistent State**: The last anonymousId is persisted to storage using the new `lastEventAnonymousId` storage key, ensuring continuity across app sessions

### Storage Integration
- **New Storage Key**: Added `Constants.storageKeys.lastEventAnonymousId` for persisting the last processed event's anonymousId
- **Rollover Trigger**: Storage rollover is automatically triggered when a new event has a different anonymousId than the previous event
- **Initialisation**: EventWriter now reads the last stored anonymousId during initialisation to maintain state across app restarts

### Event Processing Flow
1. When an event is processed, compare its anonymousId with the stored `lastEventAnonymousId`
2. If different, trigger storage rollover to finalise the current batch
3. Update the stored `lastEventAnonymousId` with the new value
4. Continue processing the event in the new batch

This approach ensures that events from different anonymous users are never mixed in the same batch, maintaining clean data separation for analytics processing.

## Checklist

- [ ] I have checked for potential performance impacts and optimized if necessary.

## How to test?

1. **Different AnonymousId Scenario**:
   - Send events with different anonymousId values / Or call RESET
   - Verify that storage creates separate batches for each unique anonymousId
   - Check that `lastEventAnonymousId` is updated in storage

2. **Same AnonymousId Scenario**:
   - Send multiple events with the same anonymousId
   - Verify that all events are grouped in a single batch
   - Ensure no unnecessary rollovers occur

3. **App Restart Scenario**:
   - Send events, restart the app, send more events
   - Verify that the last anonymousId state is preserved across sessions

## Breaking Changes

None

## Maintainers Checklist

- [ ] CI tests have passed.

## Summary by CodeRabbit

* **New Features**
  * Event batching now automatically rolls over when the anonymous user ID changes, improving per-user isolation and analytics accuracy.
  * The last seen anonymous user ID is persisted so batching remains consistent across app launches and restarts, reducing cross-user data mixing.

* **Tests**
  * Added tests verifying rollover occurs when the anonymous ID changes and that batching is preserved when the ID stays the same.

### DIFF
--- a/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
+++ b/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
@@ -102,8 +102,8 @@ final class EventWriter {
     }
 
     private func updateAnonymousIdAndRolloverIfNeeded(processingEvent: ProcessingEvent) async {
-        guard let currentEventAnonymousId = processingEvent.event?.anonymousId,
-              currentEventAnonymousId != self.lastEventAnonymousId else { return }
+        guard let lastEventAnonymousId, let currentEventAnonymousId = processingEvent.event?.anonymousId,
+              currentEventAnonymousId != lastEventAnonymousId else { return }
         
         // Rollover when last and current anonymousId are different
         await self.storage.rollover()
--- a/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
+++ b/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
@@ -17,7 +17,7 @@ final class EventWriter {
     private let writeChannel: AsyncChannel<ProcessingEvent>
     private let uploadChannel: AsyncChannel<String>
     private let flushEvent = ProcessingEvent(type: .flush)
-    private var lastEventAnonymousId: String
+    private var lastEventAnonymousId: String?
     private var storage: Storage {
         return self.analytics.configuration.storage
     }
@@ -27,8 +27,7 @@ final class EventWriter {
         self.flushPolicyFacade = FlushPolicyFacade(analytics: analytics)
         self.writeChannel = writeChannel
         self.uploadChannel = uploadChannel
-        self.lastEventAnonymousId = analytics.configuration.storage.read(key: Constants.storageKeys.lastEventAnonymousId) ??
-                                   analytics.anonymousId ?? ""
+        self.lastEventAnonymousId = storage.read(key: Constants.storageKeys.lastEventAnonymousId) ?? analytics.anonymousId
     }
     
     func put(_ event: Event) {
@@ -103,13 +102,12 @@ final class EventWriter {
     }
 
     private func updateAnonymousIdAndRolloverIfNeeded(processingEvent: ProcessingEvent) async {
-        let currentEventAnonymousId = processingEvent.event?.anonymousId ?? ""
-
-        if currentEventAnonymousId != self.lastEventAnonymousId {
-            // Rollover when last and current anonymousId are different
-            await self.storage.rollover()
-            self.lastEventAnonymousId = currentEventAnonymousId
-            self.storage.write(value: self.lastEventAnonymousId, key: Constants.storageKeys.lastEventAnonymousId)
-        }
+        guard let currentEventAnonymousId = processingEvent.event?.anonymousId,
+              currentEventAnonymousId != self.lastEventAnonymousId else { return }
+        
+        // Rollover when last and current anonymousId are different
+        await self.storage.rollover()
+        self.lastEventAnonymousId = currentEventAnonymousId
+        self.storage.write(value: self.lastEventAnonymousId, key: Constants.storageKeys.lastEventAnonymousId)
     }
 }
--- a/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
+++ b/Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
@@ -17,7 +17,7 @@ final class EventWriter {
     private let writeChannel: AsyncChannel<ProcessingEvent>
     private let uploadChannel: AsyncChannel<String>
     private let flushEvent = ProcessingEvent(type: .flush)
-    
+    private var lastEventAnonymousId: String
     private var storage: Storage {
         return self.analytics.configuration.storage
     }
@@ -27,6 +27,8 @@ final class EventWriter {
         self.flushPolicyFacade = FlushPolicyFacade(analytics: analytics)
         self.writeChannel = writeChannel
         self.uploadChannel = uploadChannel
+        self.lastEventAnonymousId = analytics.configuration.storage.read(key: Constants.storageKeys.lastEventAnonymousId) ??
+                                   analytics.anonymousId ?? ""
     }
     
     func put(_ event: Event) {
@@ -60,6 +62,8 @@ final class EventWriter {
                 
                 // Process regular events (not flush signals)
                 if !isFlushSignal {
+                    await self.updateAnonymousIdAndRolloverIfNeeded(processingEvent: event)
+
                     if let json = event.event?.jsonString {
                         LoggerAnalytics.debug(log: "Processing event: \(json)")
                         await self.storage.write(event: json)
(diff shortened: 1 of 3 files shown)


Sources/RudderStackAnalytics/Helpers/Constants.swift
```diff
@@ -70,6 +70,9 @@ struct _StorageKeys {
     /// Key for storing the anonymous user identifier.
     let anonymousId = "anonymous_id"
 
+    /// Key for storing the last event anonymous id, which is required for processing the batch of events.
+    let lastEventAnonymousId = "last_event_anonymous_id"
+
```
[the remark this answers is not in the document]
1abhishekpandey: We don't need to reset this field when those case happens. As we need to track the past value to compare with the latest value and if changed then rollover the storage!

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
```diff
@@ -27,6 +27,8 @@ final class EventWriter {
         self.flushPolicyFacade = FlushPolicyFacade(analytics: analytics)
         self.writeChannel = writeChannel
         self.uploadChannel = uploadChannel
+        self.lastEventAnonymousId = analytics.configuration.storage.read(key: Constants.storageKeys.lastEventAnonymousId) ??
+                                   analytics.anonymousId ?? ""
```
SKannaniOS: Instead of `analytics.configuration.storage` usage, use the computed property `storage`.

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
```diff
@@ -97,4 +101,15 @@ final class EventWriter {
         guard !self.writeChannel.isClosed else { return }
         self.writeChannel.close()
     }
+
+    private func updateAnonymousIdAndRolloverIfNeeded(processingEvent: ProcessingEvent) async {
+        let currentEventAnonymousId = processingEvent.event?.anonymousId ?? ""
+
+        if currentEventAnonymousId != self.lastEventAnonymousId {
```
SKannaniOS: Combine these lines using `guard let` statement, like below.

`guard let currentEventAnonymousId = processingEvent.event?.anonymousId,
                currentEventAnonymousId != self.lastEventAnonymousId else { return }`

Tests/RudderStackAnalyticsTests/EventQueue/EventQueueTests.swift
```diff
@@ -117,4 +117,82 @@ final class EventQueueTests: XCTestCase {
             await mockAnalytics.configuration.storage.remove(batchReference: item.reference)
         }
     }
+
+    func test_eventQueue_rolloverOnAnonymousIdChange() async {
+        // Given - Events with different anonymous IDs
+        let expectation = expectation(description: "Storage should rollover when anonymousId changes")
```
SKannaniOS: The test flow is fully `async/await` and runs sequentially, so the `expectation` is unnecessary. Remove it.

Tests/RudderStackAnalyticsTests/EventQueue/EventQueueTests.swift
```diff
@@ -117,4 +117,82 @@ final class EventQueueTests: XCTestCase {
...
+        // Cleanup
+        for item in finalDataItems {
+            await mockAnalytics.configuration.storage.remove(batchReference: item.reference)
+        }
+    }
+
+    func test_eventQueue_sameAnonymousIdSingleBatch() async {
+        // Given - Multiple events with the same anonymous ID
+        let expectation = expectation(description: "Events with same anonymousId should create only one batch")
+
```
SKannaniOS: The test flow is fully `async/await` and runs sequentially, so the `expectation` is unnecessary. Remove it.

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
SKannaniOS: Handle `lastEventAnonymousId` as optional variable. Don't initialize with empty string. Always there will be a value for `anonymousId`. Think about it.

1abhishekpandey: Regarding the second comment:

I generally avoid defining `optional` or `nullable` type for any variable. As that might lead to the NPE!

> Always there will be a value for anonymousId. Think about it.

When analytics is shutdown then this `analytics.anonymousId` will return `nil`. Therefore, this `?? ""` is required.

Is there any issue with the current approach?

SKannaniOS: > I generally avoid defining optional or nullable type for any variable. As that might lead to the NPE!

No. NPEs occur due to unhandled nullable variables. `Swift` addresses this at the language level with `Optionals`, which are its built-in safeguard against what other languages call a Null Pointer Exception (NPE). Swift’s syntax enforces handling of Optionals, so code with unhandled values won’t compile.

> When analytics is shutdown then this analytics.anonymousId will return nil. Therefore, this ?? "" is required.

Correct me if I’m wrong: When the SDK initializes, `UserIdentity.anonymousId` is either the previously saved value or a newly generated random UUID. Then `RudderStackDataPlanePlugin` is initialized, which in turn initializes `EventQueue`, which then initializes `EventWriter`.

Only in the initializer of `EventWriter` is the last event or current `anonymousId` (if last event anonymousId not available) collected. If neither the last event nor current `anonymousId` exists, then effectively no `anonymousId` is available—so the correct approach is to handle it as optional.

Also, after `shutdown` we don’t clear the last event `anonymousId`. That means if the SDK has already processed at least one event, the value persists. If shutdown happens before any event is processed, then no event will exist afterward—so there’s no need to default to an empty string.

1abhishekpandey: I agree that `Swift` provides proper support for `optionals`. But doesn’t the current approach work as expected? I’m trying to understand what’s wrong with it? Does it cause any issues or any unexpected behaviour? Suppose, if we proceed with the current approach do you think it'll cause any issue in future?

Regarding this:
> self.lastEventAnonymousId = analytics.configuration.storage.read(key: Constants.storageKeys.lastEventAnonymousId) ??
                                   analytics.anonymousId ?? ""

Only reason I had to add `?? ""` is because `read` and `analytics.anonymousId` is of `null` type. 

> so there’s no need to default to an empty string

This is just another approach to write the code without declaring the type as `optional`! 

I understand we could write this in an optional way, where we don't need to add this `?? ""`. But I would first like to understand what's wrong with the current approach!

I simply try to avoid `optional` or `nullable` as much as possible. Do you think it'll cause any issue, if we have the current approach?

1abhishekpandey: Also regarding your point on this:
> No. NPEs occur due to unhandled nullable variables. Swift addresses this at the language level with Optionals, which are its built-in safeguard against what other languages call a Null Pointer Exception (NPE). Swift’s syntax enforces handling of Optionals, so code with unhandled values won’t compile.

Examine this code:
```
var name: String? = nil
print(name!)
```

Suppose someone mistakenly or unintentionally write the above code, it'll lead to a runtime exception. I just don't want to have any scope for such exceptions.

**My coding style is very clear: use `nullable` or `optional` only when extremely necessary. I never try to make it my first choice.**

Please LMK if you think my approach can cause any issue and I would be happy to refactor the code.

1abhishekpandey: Based on our offline discussion, we align with the following items:
1. We are going to refactor the code to use optionals.
2. The current approach also works well without any issues.

In future, we can prefer any of the approaches: Either use an `optional` or assign the default value. Both should be fine.

Sources/RudderStackAnalytics/EventQueue/EventWriter.swift
SKannaniOS: Unwrap the `self.lastEventAnonymousId` in the same guard statement.

1abhishekpandey: I've addressed your last comment. Please review it again.

SKannaniOS (approved):

vgupta98 (approved):

PR #161, opened 2025-10-15
refactor: revamp storage module test cases

SKannaniOS:
## Description
This PR refactors the testing infrastructure for the RudderStack Swift SDK by migrating storage-related tests from XCTest to Swift Testing framework. The changes include a comprehensive MockStorage implementation that provides better testing capabilities with in-memory storage simulation and enhanced test helper methods.

## Type of change
- [ ] Documentation update

## Implementation Details
- **Added MockStorage test helper**: Created a comprehensive `MockStorage.swift` class that implements the `Storage` protocol for testing purposes
- **Migrated StorageModuleTests**: Converted all storage tests from XCTest to Swift Testing framework with improved test structure and assertions
- **Enhanced test coverage**: Added new test scenarios for storage mode validation, event batching, and helper method functionality
- **Improved test isolation**: MockStorage provides in-memory testing without side effects on disk or UserDefaults

### Key Technical Changes:
- **MockStorage Implementation**:
  - Actor-based thread-safe event storage operations
  - In-memory key-value storage with concurrent queue management
  - Support for both disk and memory storage mode simulation
  - Comprehensive test helper methods for data inspection
  - Proper Codable object encoding/decoding handling

- **Test Migration**:
  - Replaced XCTest `XCTAssert*` with Swift Testing `#expect` assertions
  - Converted test class to `@Suite` struct with `@Test` functions
  - Added parameterized tests for storage mode validation
  - Improved test naming and organization following Swift Testing conventions
  - Enhanced test readability with descriptive test names

- **Project Structure**:
  - Maintained backward compatibility for existing storage functionality

## Checklist

## How to test?
1. **Run the migrated storage tests**:
   ```bash
   xcodebuild test -workspace RudderStackAnalytics.xcworkspace -scheme RudderStackAnalytics -destination "platform=iOS Simulator,name=iPhone 15"
   ```

2. **Verify MockStorage functionality**:
   - Test key-value storage operations with various data types
   - Test event storage with batch management and rollover functionality
   - Verify thread safety with concurrent operations
   - Test storage mode simulation (disk vs memory)

3. **Validate project structure**:
   - Verify workspace builds correctly with the new test infrastructure
   - Ensure no broken references in test files

4. **Test Coverage Verification**:
   - All original test scenarios are maintained with Swift Testing
   - New test cases cover MockStorage helper methods
   - Event batching and rollover logic is thoroughly tested

## Breaking Changes
**None** - This is a test infrastructure refactoring that doesn't affect the public API or SDK functionality. The MockStorage is purely for testing purposes and doesn't impact SDK users.

## Maintainers Checklist
- [ ] CI tests have passed.

N/A - This is a backend/testing infrastructure change with no UI components.

## Additional Context
- **Context**: This work is part of SDK-4018 to investigate and modernize the test frameworks for Swift code testing
- **Swift Testing Benefits**: The migration to Swift Testing provides better parameterized testing, improved error messages, and more modern Swift syntax
- **MockStorage Advantages**: The new MockStorage implementation offers better test isolation, performance, and debugging capabilities compared to using production storage in tests

### Related Files Modified:
- `Tests/RudderStackAnalyticsTests/Helpers/MockStorage.swift` - Added (294 lines)
- `Tests/RudderStackAnalyticsTests/StorageModuleTests.swift` - Migrated to Swift Testing (306 lines modified)
- `RudderStackAnalytics.xcodeproj/project.pbxproj` - Updated project references to include MockStorage

### Testing Framework Migration Benefits:
- **Performance**: Swift Testing is generally faster than XCTest
- **Readability**: More natural Swift syntax with `#expect` instead of `XCTAssert*`
- **Parameterization**: Better support for parameterized tests with `arguments:`
- **Organization**: `@Suite` and `@Test` provide clearer test structure
- **Future-proofing**: Aligns with Apple's modern testing approach

## Summary by CodeRabbit

* **Tests**
  * Expanded and reorganized storage test suites (Memory, Disk, KeyValue, Basic) with broader coverage for batching, rollover, removal, concurrency, edge cases and special characters.
  * Added and wired a concurrency-safe in-memory mock storage into test helpers.
  * Introduced a comprehensive test harness for mock analytics, network simulation, plugin capture, logging, and parameterized analytics scenarios.
  * Removed an older legacy storage test file.


Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
```diff
@@ -5,259 +5,306 @@
...
-        
-        let value4 = true
-        storage.write(value: value4, key: "BoolValue")
-        let stored4: Bool? = storage.read(key: "BoolValue")
-        XCTAssertNotNil(stored4)
+    // MARK: - Key-Value Storage Tests
+    
+    @Test("Given MockStorage when writing primitive types then values are stored and retrievable")
+    func writePrimitiveTypes() {
+        let storage = MockStorage.forMemoryStorage()
```
1abhishekpandey: We can move this setup phase inside a different method using `@Setup` annotation. This will ensure, we don't need to write this code unnecesarily again and again in each test cases.

Objective:
- Always prefer `@Setup` to configure the **common setup phase** (and if required, then use `teardown` phase as well).

Note: If `@Setup` is not the correct annotation, please find a corresponding one for Swift!

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
```diff
@@ -5,259 +5,306 @@
...
+        let storedString: String? = storage.read(key: "StringValue")
+        #expect(storedString != nil)
+        #expect(storedString == stringValue)
+        
+        // Test Bool
+        let boolValue = true
+        storage.write(value: boolValue, key: "BoolValue")
+        let storedBool: Bool? = storage.read(key: "BoolValue")
+        #expect(storedBool != nil)
+        #expect(storedBool == boolValue)
```
1abhishekpandey: This seems a repetitive pattern. Let's focus on optimising this. 

First look if something is provided by the native Swift framework for this or not (refer [Kotlin](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/9bdbe82f854ff5c95a908c9ef59ed805bd9e31a1/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt#L799) - this is just for reference).
Or, have this:
```
    @Test
    func testStorageWriteAndRead() {
        assertStoredValue(42, key: "IntValue")
        assertStoredValue(3.14, key: "DoubleValue")
        assertStoredValue("test_string", key: "StringValue")
        assertStoredValue(true, key: "BoolValue")
    }

    private func assertStoredValue<T: Equatable>(_ value: T, key: String) {
        storage.write(value: value, key: key)
        let stored: T? = storage.read(key: key)
        #expect(stored != nil)
        #expect(stored == value)
    }
```
(Improve it if required)

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
```diff
@@ -5,259 +5,306 @@
...
-        let stored3: String? = storage.read(key: "StringValue")
-        XCTAssertNotNil(stored3)
-        
-        let value4 = true
-        storage.write(value: value4, key: "BoolValue")
-        let stored4: Bool? = storage.read(key: "BoolValue")
-        XCTAssertNotNil(stored4)
+    // MARK: - Key-Value Storage Tests
+    
+    @Test("Given MockStorage when writing primitive types then values are stored and retrievable")
```
1abhishekpandey: I would recommend to follow this pattern for writing the `given.. when.. then..` statements.

```suggestion
    @Test("Given MockStorage, when writing primitive types, then values are stored and retrievable")
```

Or,

```suggestion
    @Test("given MockStorage, when writing primitive types, then values are stored and retrievable")
```

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
```diff
@@ -5,259 +5,306 @@
...
-        await storage.rollover()
-        
-        let files = await storage.read().dataItems
-        XCTAssertFalse(files.isEmpty)
+    // MARK: - Event Storage Tests
+    
+    @Test("Given memory storage when writing events then events are batched and readable")
+    func writeAndReadEventsMemory() async {
+        let storage = MockStorage.forMemoryStorage()
+        #expect(storage.eventStorageMode == .memory)
```
1abhishekpandey: Try to avoid writing an `expect` statement even before writing the main logic. It seems to me a violation of the `given.. when.. then..` pattern. As in this case, we are directly jumping to the `then` statement.

If needed, then move this into a new test case, where you just have these two lines:
```
let storage = MockStorage.forMemoryStorage()
#expect(storage.eventStorageMode == .memory)
```

1abhishekpandey (approved):

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
SKannaniOS: - Swift’s testing framework does not have `@Setup`. If you have any references or documentation suggesting otherwise, please share them.
- The functionality of `setup()` and `tearDown()` can be mimicked using `init()` and `deinit()` when a `class` is used as a `@Suite` instead of a `struct`. However, for local instances, they are automatically deallocated at the end of the test case. Therefore, explicit `setup()` and t`earDown()` methods are generally unnecessary, unless there’s a need for shared setup steps spanning multiple lines.

1abhishekpandey: There's a requirement for the shared setup, i.e., we're using this code: `let storage = MockStorage.forMemoryStorage()` at almost all the tests. We should look forward to optimising that step!

You are free to choose the best approach. But we shouldn't repeat that line in each test.

Note: In future, use case might expand to require few other shared steps. Try to find the correct way for `@Setup` equivalent in Swift and use that to write common steps.

Regarding `@Setup`, I've already mentioned this:
> Note: If @Setup is not the correct annotation, please find a corresponding one for Swift!

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
SKannaniOS: The main benefit of using an explicit #expect is that, if the test fails, it highlights the exact line of failure. If we put the expectation inside a common method, any failure would point to that method instead, rather than the specific line in the test case.

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
SKannaniOS: #expect is a check statement that can be used anywhere. In this test case, using #expect to verify `eventStorageMode` acts like an early-exit mechanism.

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
1abhishekpandey: I understand your point. But, IMO, writing efficient code should be fine. As, I don't really expect this test case to fail and even if it fails, we can run and verify the things manually. But if you feel this is a more optimal way for you, feel free to proceed with the current approach.

BTW, have you explored if we have something similar for Swift or not?
> refer [Kotlin](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/9bdbe82f854ff5c95a908c9ef59ed805bd9e31a1/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt#L799) - this is just for reference.

As a last resort, you also have an option to move those tests into individual methods.

**The core issue with this approach, it tries to detect multiple things in a single test class and they are not even related.**

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
1abhishekpandey: >using #expect to verify eventStorageMode acts like an early-exit mechanism
If you feel, there's a need to write the verification step, move it to a separate test method - which contains only those changes. It'll achieve the same use case.

Writing a verification step, without the actual logic, is not a good thing!

SKannaniOS: Test cases revamped again. Please review again.

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
SKannaniOS: Test cases revamped again. Please review again.

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
SKannaniOS: I'll check it & implement in other modules if needed. Test cases revamped again. Please review again.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,340 @@
...
+        let value = "test_value_to_remove"
+        
+        storage.write(value: value, key: key)
+        var retrievedValue: String? = storage.read(key: key)
+        #expect(retrievedValue == value)
+        
+        storage.remove(key: key)
+        retrievedValue = storage.read(key: key)
+        #expect(retrievedValue == nil)
+    }
```
1abhishekpandey, line 79: We could write it like this:

```suggestion
    @Test("given value is stored in the storage, when key is removed, then value is deleted")
    func testKeyValueRemove() async {
        // This is the given portion - where we are storing the values
        let key = "test_remove_key"
        let value = "test_value_to_remove"
        storage.write(value: value, key: key)

        // This is the when portion, where we are removing the key
        storage.remove(key: key)

        // This is the then portion, for verifying our test
        retrievedValue = storage.read(key: key)
        #expect(retrievedValue == nil)
    }
```

1abhishekpandey: In your case, I see two issues:
1. Code is not properly strucutred i.e., `given.. when.. then..` is not aligned. Refer to my above code for structured code.
2. We unnecessarily verify something, even before performing `when` phase i.e., this line `#expect(retrievedValue == value)`.

Our aim should be to first write the `given` phase, then write the actual `when` phase and then at last write the `then` phase. We should try to avoid premature verification of anything.

Just a tip: Also, if there's no `given` phase, you may continue with the `when` and `then` phase.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,340 @@
...
+    @Test("given BasicStorage, when writing and reading string value, then value is persisted correctly")
+    func testKeyValueStringStorage() async {
+        let key = "test_string_key"
+        let value = "test_string_value"
+        
+        storage.write(value: value, key: key)
+        let retrievedValue: String? = storage.read(key: key)
+        
+        #expect(retrievedValue == value)
+    }
```
1abhishekpandey: Feel free to skip `given ..`, if we don't really have anything to write for that phase.

In this case, we could rewrite our tests like this:
```
    @Test("when writing and reading a string value, then BasicStorage persists it correctly")
    func testWhenWritingAndReadingString_thenValueIsPersisted() async {
        // When
        let key = "test_string_key"
        let value = "test_string_value"
        storage.write(value: value, key: key)
        let retrievedValue: String? = storage.read(key: key)
    
        // Then
        #expect(retrievedValue == value)
    }
```

1abhishekpandey: Note: If you write `given ..`, then you need to show that in tests as well. 

Currently, we write this `given BasicStorage`. But I cannot see anything related to that in the tests.

Let's follow the same approach in other tests.

Tests/RudderStackAnalyticsTests/StorageModuleTests/DiskStoreTests.swift
```diff
@@ -0,0 +1,405 @@
...
+    
+    deinit {
+        keyValueStore.removeAll()
+    }
+    
+    // MARK: - Initialization Tests
+    
+    @Test("given writeKey, when initializing DiskStore, then store is created with correct writeKey and storage URL")
+    func testInitialization() async {
+        #expect(await store.writeKey == testWriteKey)
```
1abhishekpandey: Let's remove this expect statement.

If you really have a need to verify such thing, move it into a new test.

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
1abhishekpandey: We've decided to refactor the code.

Tests/RudderStackAnalyticsTests/StorageModuleTests.swift
1abhishekpandey: We've decided to refactor the code.

Tests/RudderStackAnalyticsTests/StorageModuleTests/KeyValueStoreTests.swift
```diff
@@ -0,0 +1,447 @@
...
+    @Test("given writeKey, when initializing KeyValueStore, then store is created successfully")
+    func testInitialization() async {
+        let writeKey = "test-init-key"
+        let keyValueStore = KeyValueStore(writeKey: writeKey)
+        
+        // Verify store can be used after initialization
+        keyValueStore.save(value: "test", reference: "init_test")
+        let result: String? = keyValueStore.read(reference: "init_test")
+        #expect(result == "test")
+    }
```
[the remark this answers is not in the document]
SKannaniOS: We have deinit implementation.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
SKannaniOS: The `storage` itself the instance of `BasicStorage` only. So given `BasicStorage` is fine..

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
SKannaniOS: I commented the test case with steps.

Tests/RudderStackAnalyticsTests/Helpers/SwiftTestMockProvider.swift
```diff
@@ -0,0 +1,503 @@
...
+                client?.urlProtocol(self, didFailWithError: URLError(.badURL))
+                return
+            }
+            
+            let response = HTTPURLResponse(
+                url: url,
+                statusCode: statusCode,
+                httpVersion: "HTTP/1.1",
+                headerFields: headers
+            )!
```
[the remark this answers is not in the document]
SKannaniOS: Will update this class in future PRs.

Tests/RudderStackAnalyticsTests/Helpers/SwiftTestMockProvider.swift
```diff
@@ -0,0 +1,539 @@
...
+    
+    func runAfter(_ seconds: Double, block: @escaping () async -> Void) async {
+        // Suspend the current task for the specified duration
+        let nanoseconds = UInt64(seconds * 1_000_000_000)
+        try? await Task.sleep(nanoseconds: nanoseconds)
+        
+        // Execute the block after the delay
+        await block()
+    }
+}
```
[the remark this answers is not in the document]
SKannaniOS: Will be revamped in future PRs.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,348 @@
...
+        """
+    var storage: BasicStorage
+    
+    init() {
+        storage = BasicStorage(writeKey: testWriteKey, storageMode: .memory)
+    }
+    
+    // MARK: - KeyValueStorage Tests
+    
+    @Test("given BasicStorage, when writing and reading string value, then value is persisted correctly")
```
saikumarrs:
```suggestion
    @Test("given BasicStorage, when writing string value, then value is persisted correctly")
```

Tests/RudderStackAnalyticsTests/StorageModuleTests/DiskStoreTests.swift
```diff
@@ -0,0 +1,406 @@
...
+    func testStoreSingleEvent() async {
+        await store.retain(value: sampleEventJson)
+        
+        let retrievedItems = await store.retrieve()
+        #expect(retrievedItems.isEmpty) // Should be empty as we haven't rolled over yet
+        
+        // Check if file exists on disk but isn't marked as complete
+        let fileStorageURL = await store.fileStorageURL
+        let contents = FileManager.contentsOf(directory: fileStorageURL.path)
+        #expect(contents.count >= 1) // Should have at least one .tmp file
```
saikumarrs: This does not clearly validate that the correct file exists.

Tests/RudderStackAnalyticsTests/StorageModuleTests/DiskStoreTests.swift
```diff
@@ -0,0 +1,406 @@
...
+        
+        // Clean up the store
+        await store.removeAll()
```
saikumarrs, line 56: Similar to init and deinit, do we not have any provision to clear the state for each test case?

Tests/RudderStackAnalyticsTests/StorageModuleTests/DiskStoreTests.swift
SKannaniOS: This test checks file created or not. I'll update the validation.

Tests/RudderStackAnalyticsTests/StorageModuleTests/DiskStoreTests.swift
SKannaniOS: There is no exact function for that. Either we need to use deinit() or create one function and call it at the end of each test case.

saikumarrs: Ok. For other test suite where the operation is more than one statement, let's create a function and re-use it.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
saikumarrs: @SKannaniOS, I just pointed out test case assuming you'd fix others too 😄

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,348 @@
...
+    @Test("given BasicStorage, when writing string value, then value is persisted correctly")
+    func testKeyValueStringStorage() async {
+        let key = "test_string_key"
+        let value = "test_string_value"
+        
+        storage.write(value: value, key: key)
+        let retrievedValue: String? = storage.read(key: key)
+        
+        #expect(retrievedValue == value)
+    }
```
1abhishekpandey: Let's use parametrised testing for all these test cases!

(I can see we've already implemented it at other test files, let's implement that over here.)

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,348 @@
...
+        // Store event data
+        await storage.write(event: sampleEventJson)
+        await storage.rollover()
+        
+        // Verify data exists
+        let keyValue: String? = storage.read(key: "test_key")
+        #expect(keyValue == "test_value")
+        
+        let eventResult = await storage.read()
+        #expect(eventResult.dataItems.count == 1)
```
1abhishekpandey: Let's try to avoid verifying anything before we write the "when" phase.

1abhishekpandey: Our aim should be to align our test cases with the `given.. when.. then..` statement. 

Removing the early `verify` statement makes our test cleaner. 

As said earlier, if you feel they are needed, feel free to add a new test case with those lines.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,348 @@
...
+            """
+        
+        await storage.write(event: specialEvent)
+        await storage.rollover()
+        
+        let result = await storage.read()
+        #expect(result.dataItems.count == 1)
+        
+        let batchContent = result.dataItems.first?.batch ?? ""
+        #expect(batchContent.contains("🚀") == true)
```
1abhishekpandey, line 294: Does the batch really contain this icon?

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
```diff
@@ -0,0 +1,348 @@
...
+    @Test("given BasicStorage with stored value, when removing key, then value is deleted")
+    func testKeyValueRemove() async {
+        // given...
+        let key = "test_remove_key"
+        let value = "test_value_to_remove"
+        storage.write(value: value, key: key)
+        
+        // confirm value is stored..
+        var retrievedValue: String? = storage.read(key: key)
+        #expect(retrievedValue == value)
```
1abhishekpandey: Let's optimise this by removing this early verification!

Tests/RudderStackAnalyticsTests/AnalyticsTests.swift
```diff
@@ -0,0 +1,842 @@
...
+        analytics.isAnalyticsActive = true
+        
+        mockPlugin = MockEventCapturePlugin()
+        analytics.add(plugin: mockPlugin)
+    }
+    
+    // MARK: - Track Event Tests
+    @Test("given Analytics, when tracking events with variations, then event is captured correctly", arguments: EventTestCaseParameters.trackEvent)
+    func testTrackEventVariants(_ eventName: String, _ properties: Properties?, _ options: RudderOption?) async {
+        // when
```
1abhishekpandey: We shouldn't write this `when` line!
```suggestion
```

Tests/RudderStackAnalyticsTests/AnalyticsTests.swift
```diff
@@ -0,0 +1,842 @@
...
+        
+        // Validate properties if provided
+        if let properties {
+            let raw = event.properties?.dictionary?.rawDictionary ?? [:]
+            for (key, value) in properties {
+                #expect("\(raw[key] ?? "")" == "\(value)")
+            }
+        }
+        
+        // Validate options if provided
```
1abhishekpandey, line 64: Let's try to define a single function to extract out the common logic. This will make the test more readable.

We actively used this pattern in coding to avoid repating the code twice (for reference - refer [here](https://github.com/rudderlabs/rudder-sdk-kotlin/blob/9bdbe82f854ff5c95a908c9ef59ed805bd9e31a1/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt#L342-L345))!

If you think some codes are being repeated across test files, let's create a different Utils file to move the common logic over there.

Our objective should be to write optimal code which are easy to maintain.

Current issue:
- If you need to make changes in a single place, (say) in the `integration` logic - you need to make changes at least at five different places.
- Also, they don't look clean, and since they're repetitive and very long code they're very hard to review as well.

1abhishekpandey: Let's try to optimise this across different test methods!

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
1abhishekpandey: @saikumarrs I still believe we are not utilising the `given ...` pattern properly. The current approach used in this PR doesn't seem to help a lot!

As per my understanding, we should use `given ..` to write exactly what we are assuming. This will provide more context for the test. When we are just writing `given <ClassName>`, we seem to lose the crucial piece of information, which could help in maintaining the test and also improve the readability.

For ex: Let's say we are testing the `SourceConfigManager` class and we want to simulate the network condition logic where the status code would be 400. Then in that case, we can write the statement like this: `given network returns 400, when .., then ..`. But current way of writing does not seem an optimal way to me: `given SourceConfigManager ..., when .., then ..`.

Repeating `given <ClassName>` seem redundant to me.

For a better example, like how one should use the `given ..` pattern, we should refer to the Kotlin: https://github.com/rudderlabs/rudder-sdk-kotlin/blob/develop/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/SourceConfigManagerTest.kt and https://github.com/rudderlabs/rudder-sdk-kotlin/blob/develop/core/src/test/kotlin/com/rudderstack/sdk/kotlin/core/AnalyticsTest.kt (for references). 

If we still believe, the current approach used in this PR is fine, then we can proceed with it.
CC: @SKannaniOS

saikumarrs: @SKannaniOS, let's address the test case descriptions wherever possible to be more meaningful.

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
SKannaniOS: This test case is about how the SDK stores the event json which has some special characters. So this test case needed.

1abhishekpandey (approved):

saikumarrs (approved):

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
saikumarrs: @1abhishekpandey, is this still unresolved?

1abhishekpandey: It's not fully addressed. Therefore, I've kept it open.

I think we can proceed without it, as I don't want to block this PR for this small change.

saikumarrs: @SKannaniOS lrt's align with Abhishek to either defer or address it now.

SKannaniOS: I'll update all the test descriptions again & push the changes.

1abhishekpandey (approved):

Tests/RudderStackAnalyticsTests/StorageModuleTests/BasicStorageTests.swift
1abhishekpandey: It looks better. Let's proceed with this.

PR #202, opened 2026-02-05
fix(objc): remove nil-coalescing fallback in ObjCPluginAdapter intercept method

SKannaniOS:
## Description
This PR fixes the event interception logic in the `ObjCPluginAdapter` by removing the nil-coalescing fallback (`?? event`) in the intercept method. Previously, when an ObjC plugin's intercept method returned nil, the original event would be returned as a fallback. This change ensures that when a plugin explicitly returns nil from the intercept method, it properly signals that the event should be filtered out/dropped, allowing for proper event filtering behavior.

The change maintains the intended plugin architecture where plugins can choose to filter out events by returning nil, rather than always falling back to the original event.

## Type of change

- [ ] Documentation update

## Implementation Details

- Modified `ObjCPluginAdapter.intercept(event:)` method in `Sources/RudderStackAnalytics/ObjC/Plugins/Base/ObjCPlugin.swift`
- Removed the nil-coalescing operator (`?? event`) from line 94
- Changed return statement from `objcPlugin.intercept?(objcEvent)?.event ?? event` to `objcPlugin.intercept?(objcEvent)?.event`
- This allows plugins to properly filter out events by returning nil instead of always falling back to the original event

## Checklist

- [ ] I have added the necessary documentation (if appropriate).

## How to test?

To test this change:

1. **Create a test ObjC plugin** that implements the intercept method and returns nil for certain events
2. **Verify event filtering**: Ensure that when the plugin returns nil, the event is properly filtered out (not sent to downstream plugins or tracked)
3. **Verify normal operation**: Ensure that when the plugin returns a modified event, it continues to work as expected
4. **Unit tests**: Add tests for the ObjCPluginAdapter class to verify both scenarios:
   - When `intercept` returns nil (event should be filtered)
   - When `intercept` returns a valid event (event should be passed through)

## Breaking Changes

This could be considered a breaking change for existing ObjC plugins that rely on the previous fallback behavior. 

## Maintainers Checklist

- [ ] CI tests have passed.

N/A - This is a backend logic change with no UI impact.

## Additional Context

This change aligns the ObjC plugin adapter behavior with the expected plugin architecture where returning nil from an intercept method should filter out the event rather than falling back to the original event. This is consistent with how other plugin adapters and the broader plugin system is designed to work.

The change is minimal and focused, affecting only the specific return logic in the ObjCPluginAdapter class. This ensures that plugin developers have full control over event filtering behavior when using ObjC-based plugins.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed event handling in the Objective-C plugin adapter so events are now dropped when interception yields no result instead of being forwarded unchanged. This corrects prior unexpected behavior and ensures intercepted events that indicate they should be suppressed are consistently filtered out, improving reliability of event processing without changing any public API signatures.

### DIFF
--- a/Sources/RudderStackAnalytics/ObjC/Plugins/Base/ObjCPlugin.swift
+++ b/Sources/RudderStackAnalytics/ObjC/Plugins/Base/ObjCPlugin.swift
@@ -91,7 +91,7 @@ final class ObjCPluginAdapter: Plugin {
      */
     func intercept(event: Event) -> Event? {
         let objcEvent = ObjCEvent.createObjCEvent(from: event)
-        return objcPlugin.intercept?(objcEvent)?.event ?? event
+        return objcPlugin.intercept?(objcEvent)?.event
     }
 
     /**
--- a/Sources/RudderStackAnalytics/ObjC/Plugins/Base/ObjCPlugin.swift
+++ b/Sources/RudderStackAnalytics/ObjC/Plugins/Base/ObjCPlugin.swift
@@ -91,7 +91,10 @@ final class ObjCPluginAdapter: Plugin {
      */
     func intercept(event: Event) -> Event? {
         let objcEvent = ObjCEvent.createObjCEvent(from: event)
-        return objcPlugin.intercept?(objcEvent)?.event
+        guard let intercept = objcPlugin.intercept else {
+            return event
+        }
+        return intercept(objcEvent)?.event
     }
 
     /**


vgupta98 (approved):

vgupta98:
Lets have some tests for the ObjCPlugin as well.

SKannaniOS:
@vgupta98 
We’ll plan to add test cases in the near future only for the Objective-C wrapper classes that differ from or contain additional implementations beyond the underlying Swift classes.

vgupta98 (approved):

ChryssaAliferi (approved):

